Issue #28, opened 2016-02-10
Switch to supported Anaconda env


pllim:
When the time comes to package this with standard STScI software distribution.

pllim:
Closed by #30

Issue #29, opened 2016-02-11
Do we need ui.qt.uic in the repo?


pllim:
I am not very sure what `pyfocal.ui.qt.uic` is supposed to do in the package.

pllim:
Somewhat related: The following files in the repo claims to be generated by something else, so do they really need to be version controlled if they can be generated on-the-fly?

* `ui/qt/icon_resource_rc.py`
* `ui/qt/mainwindow.py`
* `ui/qt/plotsubwindow.py`

nmearl:
The uic directory contains the sources for the qt files used with the Qt Creator program to generate gui objects. I want these versioned, so, yes, I very much want to keep them in the repo.

> The following files in the repo claims to be generated by something else

These are the results of the conversion from the source files to python. These are what are actually used when you launch the application. They are necessary.

pllim:
Okay.

PR #30, opened 2016-02-11
Make pyfocal able to run in Anaconda Python 2.7

### LINKED ISSUE #24: ImportError qtpy.QtWidgets
Under Anaconda, I am unable to start `pyfocal`. Am I missing some dependency? `qtpy` is nothing I have heard of before.

```python
  File ".../bin/pyfocal", line 9, in <module>
    load_entry_point('pyfocal==0.1a1', 'console_scripts', 'pyfocal')()
  File ".../lib/python2.7/site-packages/setuptools-19.6.2-py2.7.egg/pkg_resources/__init__.py", line 547, in load_entry_point
  File ".../lib/python2.7/site-packages/setuptools-19.6.2-py2.7.egg/pkg_resources/__init__.py", line 2719, in load_entry_point
  File ".../lib/python2.7/site-packages/setuptools-19.6.2-py2.7.egg/pkg_resources/__init__.py", line 2379, in load
  File ".../lib/python2.7/site-packages/setuptools-19.6.2-py2.7.egg/pkg_resources/__init__.py", line 2385, in resolve
  File ".../site-packages/pyfocal-0.1a1-py2.7.egg/pyfocal/app.py", line 7, in <module>
    from qtpy.QtWidgets import *
ImportError: No module named qtpy.QtWidgets
```

pllim:
This might be a bit contraversial, but it enables `pyfocal` to be run from Anaconda in Python 2. Effectively fixes #24 and #28. 

Here is what I see without using Homebrew etc.

screenshot


pllim:
@nmearl , what do you think? I understand if you want to freeze your current master and hold this off. We can also create a tag for the version freeze and merge this into master. Up to you.

c/c @ibusko , @stscieisenhamer

nmearl:
This looks good. One thing to note is that any file in the `ui/qt` directory will be overwritten when they are generated again, so I would strongly suggest not editing this files. I realize this conflicts because they natively import PyQt5 right now.

The solution is to extend the bash file in `ui/qt/uic/gen_uic.sh` to edit the python files after generation to replace those lines with calls to `qtpy`.

pllim:
Re: uic

That is why I asked about versioning them in #29. Why not just have `pyfocal` generate it upon install if it depends on what user has? Or pre-generate them for all supported Qt4/Qt5/PySide and then choose during runtime?

nmearl:
Generating the python files requires the user to have those qt tools installed on their system, which will most likely not be the case.

Pre-generation is probably the best bet, and the one I was initially aiming for (essentially an expansion of how it currently is implemented).

pllim:
Anyways, I think we should handle the `uic` stuff as a separate issue. Please test this PR and make sure I didn't break anything on your side. And merge however you want it, if you wish.

nmearl:
:+1: Locally. I'm going to merge this now, and we'll use this as the "frozen" distribution.

pllim:
Thanks!

Issue #19, opened 2016-02-08
Data loader for tabular files


nmearl:
Create a registry object for ingesting data files that are e.g. `csv`, `txt`, etc. This will require either

1. Creating a new `YAMLLoader` class to deal with non-fits styled data, or
2. Extending the current `YAMLLoader` to also deal with the tabular data.

See [here](https://github.com/nmearl/pyfocal/blob/1aa0bc4645f76d94e2d34ce64ae9d92998e431ef/pyfocal/interfaces/registries.py#L35) for details on current implementation.

pllim:
This is a bit too cryptic. Exactly what kind of file formats are you expecting? Is it possible to use any existing parser, say, from JWST pipeline? We have to be careful not to re-invent the wheel here.

nmearl:
Okay. Well, since we're generally leveraging the astropy library for most all our functionality, I think we ought to stay within that realm.

For instance, tabular data is easily read into an [astropy table object](http://docs.astropy.org/en/stable/api/astropy.table.Table.html). From there, it's possible to export the normalized object as an ndarray that can be used in the creation of the `Data` object we use in Pyfocal.

The associated yaml file should contain definitions for how to ingest that data, e.g.

```yaml
--- !CustomLoader
name: Generic Table
extension: [csv, txt]
data:
  col: 0
  unit: "Jy"
dispersion:
  col: 1
  unit: "Angstrom"
uncertainty:
  col: 2
meta:
  author: "Nicholas Earl"
```

pllim:
That's fine for now, but keep in mind that sooner or later, this tool will have to support JWST data products.

nmearl:
> this tool will have to support JWST data products

I'm not sure what you mean by this. The idea is to make loaders generic enough that either we can include ones ourselves in the registries for any standardized jwst data, or have users easily create their own.

pllim:
So, let's say I have my calibrated spectrum file from JWST or HST pipeline, how do I ingest it into `pyfocal`? Maybe it is just a matter of documentation.

nmearl:
> Maybe it is just a matter of documentation.

[...]

PR #42, opened 2016-02-12
Added ASCII loader

pllim:
Added ASCII loader (see #19). Added user-defined dispersion unit to `Data` (spectrum object).


pllim:
This is tested to work with basic and IPAC ASCII formats for the "generic" test data for the sprint.

PR #58, opened 2016-02-16
Add "Known Issues" section with workaround for qt4/qt5 error

eteq:
This adds a "Known Issues" section to the installation docs with a quick fix for the "this application failed to start because it could not find or load the Qt platform plugin "cocoa"." error


nmearl:
Awesome! :+1:

PR #77, opened 2016-02-17
Fixed doc intersphinx linking and added arith text

pllim:
Fixed doc intersphinx linking and added arith text

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,9 +15,9 @@
 
 import sys
 import os
-import sphinx_rtd_theme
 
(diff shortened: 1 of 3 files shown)


eteq:
@pllim - are you sure the ``default_role = 'py:obj'`` part is necessary? I thought that I'd the default default role?

(No reason *not* to, anyway - mostly my own curiosity)

pllim:
@eteq , if I don't set that, I need to explicitly specify `:class:` role for it to link properly. Not sure why, because I didn't have to in the past. I am using Sphinx 1.3.1.

PR #80, opened 2016-02-17
Adding ability to write and read models to/from files.

ibusko:
This is partial work (only writing to file is implemented so far) and involves changes in the UI (new Save and Read buttons in the model editor window), hence the pull request. 

My goal at this point is more in the line of a code review than an actual code merge. This is work that was partially implemented in the old specview but never made the light of the day. The model IO code comes mostly from the https://github.com/ibusko/specfit project. Some of the code is outdated and should be eventually rewritten, but is OK as a staring point.


pllim:
Some accompanying user-facing documentation would be desirable. Doesn't have to be fancy, just need enough info so user can use the new feature.

nmearl:
I agree with @pllim. Perhaps in the `pyfocal/docs/source/model_fitting.rst` file.

Also, go ahead and run the `gen_rc.sh` in `pyfocal/ui/qt/uic/` (first `cd` to that directory).

```
sh gen_uic.sh pyqt5
```

nmearl:
Once you run the `gen_uic.sh` file, we can pull this in

pllim:
Also needs rebasing.

ibusko:
I run gen_ui, tested the resulting code base, and rebased. 

The one thing I didn't succeed is the layout for the two new buttons Save and Read. The buttons look OK when running the QT Designer, but in the running app they are partially covered by the widow border, and the Save button has its text garbled when the app starts. Once it is used, the text fixes itself.

pllim:
@ibusko , a screenshot here would help, as we might not see the exact same thing you see.

ibusko:
I am closing this pull request and starting a new one. I resolved the conflicts but github doesn't seem to see that I did.

PR #99, opened 2016-02-19
Layer arithmetic feature

nmearl:
Layer arithmetic has been implemented. This includes some new icons (and some not connected).

Also, there is a bug in the current release version of Astropy with NDData arithmetic. It's fixed in the current master branch of Astropy, and (after awhile of head->wall) I've implemented a simply work around for now. Perhaps @eteq can give an ETA on when the next Astropy release will occur?


nmearl:
Okay, nevermind. The workarounds are not robust enough, and I don't want to have to refactor our code to trick astropy into doing the right thing. Maybe we should wait until astropy gets update, or use the develop version of astropy with pyfocal.

nmearl:
I've disabled arithmetic so I can merge other changes within this PR.

pllim:
Since Pyfocal is still in development stage, it is okay to require user to use dev version of Astropy, right?

PR #107, opened 2016-02-20
New: Adding ability to write and read models to/from files.

ibusko:
Still says that cannot automatically merge, although there're no conflicts.


ibusko:
Conflicts are coming from the ui files. It seems that github doesn't deal with those in a graceful way. I will roll back my changes to the ui files, rebase and merge my fork, and then do the ui changes directly on the origin repo.

PR #166, opened 2016-02-26
Add package data for doc

### LINKED ISSUE #136: How to handle flux unit electrons/s
Some data has
```python
BUNIT = 'ELECTRONS/S'
```
but Astropy does not recognize it. `specviz` needs to auto-magically correct that somehow.

pllim:
First part of custom loader doc improvement, as discussed with Susan and Gisella. They want data files to go with our YAML files, so user can use them as examples to make their own.

Also fixes #136.

I'll add the actual doc change in a separate PR.


pllim:
Note that I did not change setup to actually install the data files. I don't see the need because they only serve as documentation examples. But we can always update setup script in the future if there is a need to actually install them.

Issue #100, opened 2016-02-19
Astropy's NDData arthmetic buggy in current release


nmearl:
The arithmetic mixin and nddata objects have several bugs that are fixed in the development branch of astropy, but not in the release.

I previously had several workarounds, but they were not robust enough to prevent errors. I've disabled the arithmetic for now. I'm pasting a reproducible bug error below for the current version of astropy (I had began to write a PR, but realized they fixed this in the development branch).

``` python
class TestData(NDData):
    def __init__(*args, **kwargs):
        super(TestData, self).__init__(*args, **kwargs)

a = NewData(np.random.sample(10), uncertainty=StdDevUncertainty(np.random.sample(10)*0.1))
b = NewData(np.random.sample(10), uncertainty=StdDevUncertainty(np.random.sample(10)*0.1))
c = a.multiply(b)

/Users/nearl/local/lib/python2.7/site-packages/astropy/nddata/mixins/ndarithmetic.pyc in _arithmetic(self, operand, propagate_uncertainties, name, operation)
/Users/nearl/local/lib/python2.7/site-packages/astropy/nddata/nduncertainty.pyc in propagate_multiply(self, other_nddata, result_data)
/Users/nearl/local/lib/python2.7/site-packages/astropy/nddata/nduncertainty.pyc in parent_nddata(self)
MissingDataAssociationException: Uncertainty is not associated with an NDData object
```

nmearl:
As @pllim pointed out, it's alright if we use development version of astropy for now. I've re-enabled arithmetic, and added the dev version to the install requirements in d606e8f794a66b540a53655a1f125964abe602b6.

pllim:
Might want to note that in user-facing documentation about installation too.

eteq:
:+1: to making sure users know that they need the dev version.

@pllim - are these fixes to actual bugs?  I.e., a 1.1.2 release will fix them?  Or "buggy" in the sense of some substantial behavior change that has to wait for 1.2?

eteq:
(In case it's not clear, the relevance of that question is that a 1.1.2 release can occur very soon, like probably next week.  But 1.2 isn't until the summer)

pllim:
@eteq , not sure. I am not following Astropy NDData development very closely. But given that Pyfocal is likely to remain in alpha (or maybe beta?) stage in summer anyway, I don't think this is a huge issue since we have a workaround by using dev.

What do you think, @nmearl ?

eteq:
[...]

PR #192, opened 2016-03-17
Change astropy to release version

eteq:
This changes the requirements.txt file to call for using the release version instead of ``nmearl/astropy``. We can put in a specific version if that's better, but if there was a specific reason why specviz needs the *dev* version of astropy, I can instead modify this to point to ``astropy/astropy`` instead of ``nmearl/astropy``.  Better yet, if we can point to a specific *commit*, that's even better.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@
 
 jinja2
 cython
-astropy
+git+https://github.com/astropy/astropy.git@ffc0a89b2c42fd440eb19bcb2f93db90cab3c98b
 git+https://github.com/nmearl/pyqtgraph.git
 -e .
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@
 
 jinja2
 cython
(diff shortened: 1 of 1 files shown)


pllim:
@eteq , definitely need the dev version for now. See discussions in #100.

eteq:
Ooh, right, gotcha.  How about the version I just pushed up, then?  It points to the main astropy repo instead of nmearl/astropy, and a specific commit.  Or @nmearl, was there a specific reason you're using *your* master?

Presumably this could then just be updated when astropy v1.2 comes out.

pllim:
:+1: on using specific commit directly from `astropy` but I'll let @nmearl have the last word.

nmearl:
> It points to the main astropy repo instead of nmearl/astropy, and a specific commit.

That's perfect. I was just using my fork as temporary "frozen version" of astropy. Pointing to the specific commit achieves the same end in a more direct way. I'm all for it.

eteq:
:+1:

PR #193, opened 2016-03-17
Update conda install instructions to be better-aligned to specviz name

eteq:
I think these instructions were written when the name was ``pyfocal`` - this just changes the name used for the conda environment to be more like ``specviz``.


pllim:
:+1:

PR #201, opened 2016-05-25
Line identifications

ibusko:
This pull request adds the ability to plot on screen line identification markers. The functionality is very basic, mostly appropriate for demoing the concept. Basically, there is a new menu entry (under the gear menu button) that, wen clicked, reads from pre-canned line lists and plots identifiers for all lines that fall within the wavelength range of the data. Much work is still needed, especially in adding features to help the ingestion, editing, and combination of spectral line lists, including user-supplied lists and lists provided by VO line list servers.

I think some ideas can be gotten from the Java specview:  http://specview.stsci.edu/tutorials/tutlines.pdf


specviz/app.py
```diff
@@ -71,5 +71,7 @@ def sigint_handler(*args):
     QApplication.quit()
 
 
+# test commit
```
pllim: This should  be removed.

pllim:
Is it necessary to create a custom loader for line list, which looks like simple ASCII table? Can't we just use `astropy.io.ascii` to parse it?

pllim:
Maybe use `'ecsv'` format, as defined by https://github.com/astropy/astropy-APEs/blob/master/APE6.rst ?

nmearl:
> s it necessary to create a custom loader for line list, which looks like simple ASCII table? Can't we just use astropy.io.ascii to parse it?

The custom loader is just a proxy to Astropy's ascii reader (see [here](https://github.com/spacetelescope/specviz/pull/201/files#diff-a19edea551905479118517b248ad785fR398)). But including a custom loader, a user can just click the folder icon to load the line list.

It seems that's what Ivo is going for?

pllim:
Okay.

specviz/app.py
ibusko: Done.

ibusko:
The custom loader in its current form is just is a place holder for a more complex interface that should be able to read *any* format, not just ASCII. Line lists come in a large variety of shapes, formats, and contents, including VO tables and other exotic forms delivered by, say, line list servers (SLAP protocol). Besides, I know from past experience that the very first user request for improvement would be to be able to ingest this funny-looking, non-standard user-supplied line list. 

Even with plain ASCII tables, there is the need to be able to read all columns from any table, present then to the user, and allow the user to interact with that list-specific information. For instance, specviz should know if a given table column should be sortable in numerical or alpha-numerical order. A YAML file descriptor is the place to put all that information. See the interface on Section 2 of that tutorial to have an idea of what I am talking about.

ibusko:
For instance, this is a typical line list:

http://specview.cvs.sourceforge.net/viewvc/specview/resources/linelists/illss.txt?revision=1.1&view=markup

and this is the associated descriptor in Java specview:

http://specview.cvs.sourceforge.net/viewvc/specview/resources/linelists/illss.xml?view=markup

specviz/ui/widgets/sub_windows.py
```diff
@@ -264,8 +333,8 @@ def add_container(self, container=None, window=None):
         else:
             container.change_units(*self._plot_units)
 
-        if container.error is not None:
-            self._plot_item.addItem(container.error)
+        # if container.error is not None:
```
ibusko: These should be back in the code.

PR #203, opened 2016-06-07
Added missing entry for linelists in package_data

astrofrog:
Without this, specviz doesn't work when doing a clean install


pllim:
I am just going to merge this as this is the right way to go.

PR #217, opened 2016-10-10
Convert readthedocs links for their .org -> .io migration for hosted projects

adamchainz:
As per [their blog post of the 27th April](https://blog.readthedocs.com/securing-subdomains/) ‘Securing subdomains’:

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.

Test Plan: Manually visited all the links I’ve modified.

### DIFF
--- a/docs/source/model_fitting.rst
+++ b/docs/source/model_fitting.rst
@@ -4,7 +4,7 @@ Model Fitting
 =============
 
 SpecViz utilizes
-`Astropy Models and Fitting <http://astropy.readthedocs.org/en/latest/modeling/index.html>`_
+:ref:`Astropy Models and Fitting <astropy:astropy-modeling>`
 to fit models to its spectra. For example, you can fit one model to the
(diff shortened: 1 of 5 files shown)


docs/source/model_fitting.rst
```diff
@@ -4,7 +4,7 @@ Model Fitting
 =============
 
 SpecViz utilizes
-`Astropy Models and Fitting <http://astropy.readthedocs.org/en/latest/modeling/index.html>`_
+`Astropy Models and Fitting <https://astropy.readthedocs.io/en/latest/modeling/index.html>`_
```
pllim: Thanks for doing this! While you are at this, would you mind changing this to use intersphinx instead? That is, link it like this:
```
:ref:`Astropy Models and Fitting <astropy:astropy-modeling>`
```

pllim:
Thanks!

Issue #219, opened 2016-10-28
Override units in custom loader


hcferguson:
The VIPERS project distributed FITS files with the spectra in a FITS binary table, which can be read successfully with Table.read() but which has non-standard units. I created a custom loader, but it fails I think because the fits reader itself rejects the units before specviz can re-interpret them. 

```
In [15]: t = Table.read('VIPERS_110055275_corr.fits')

In [16]: print t
 WAVES        FLUXES         NOISE        SKY        EDIT      MASK
Angstrom erg / (A cm2 s) erg/s/cm^2/Ang  counts erg/s/cm^2/Ang     
-------- --------------- -------------- ------- -------------- ----
 5514.28    -1.28638e-18    1.89357e-18 81.5747    1.64353e-18    1
 5521.42     1.48216e-18    1.60909e-18 94.9008    1.63095e-18    1
```

Here's my custom loader:

```
more ~/.specviz/vipers.yaml
--- !CustomLoader
name: VIPERS
extension: [fits]
wcs:
  hdu: 0
dispersion:
  hdu: 1
  col: 0
  unit: 'Angstrom'
data:
  hdu: 1
  col: 1
  unit: 'erg s-1 cm-2 Angstrom-1'
uncertainty:
  hdu: 1
  col: 2
  type: 'std'
  unit: 'erg s-1 cm-2 Angstrom-1'
meta:
  author: H. Ferguson
```

Here's the traceback:

```
INFO:root:Attempting to open '/Users/hcferguson/Documents/conferences/2016/montreal/presentation/VIPERS/VIPERS_110055275.fits' using filter '<class 'filter'>'.
WARNING: FITSFixedWarning: The WCS transformation has more axes (2) than the image it is associated with (0) [astropy.wcs.wcs]
WARNING:astropy:FITSFixedWarning: The WCS transformation has more axes (2) than the image it is associated with (0)
WARNING: UnitsWarning: The unit 'angstrom' has been deprecated in the FITS standard. Suggested: 10**-1 nm. [astropy.units.format.utils]
WARNING:astropy:UnitsWarning: The unit 'angstrom' has been deprecated in the FITS standard. Suggested: 10**-1 nm.
WARNING: UnitsWarning: The unit 'erg' has been deprecated in the FITS standard. Suggested: cm2 g s-2. [astropy.units.format.utils]
WARNING:astropy:UnitsWarning: The unit 'erg' has been deprecated in the FITS standard. Suggested: cm2 g s-2.
[...]

PR #262, opened 2017-03-08
Issue#219

mattjhill:
Change flux unit determination logic to
1. YAML loader unit
2. table column unit
3. FITS header unit


mattjhill:
This addresses #219

PR #269, opened 2017-03-20
replace GaussianAbsorption1D with GaussianAbsorption

ibusko:
This fixes the problem of disappearing absorption lines when fitting. Uses a better definition for an absorption line. The definition in astropy seems questionable.

This is an example of a fit using the new model:

absorption


ibusko:
Please disregard this pull request. It is superseded by "Better definition for an absorption Gaussian #271"

PR #271, opened 2017-03-24
Better definition for an absorption Gaussian - remove RedshiftScaleFactor

ibusko:
This code supersedes the former pull request. Please disregard that one.

### DIFF
--- a/specviz/analysis/models/gaussian_absorption.py
+++ b/specviz/analysis/models/gaussian_absorption.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from astropy.modeling import Fittable1DModel, models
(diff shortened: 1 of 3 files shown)


nmearl:
What's the difference between `GaussianAbsorption` and `Gaussian1D`?

ibusko:
Ivo answers:

The `GaussianAbsorption` model is  the `Gaussian1D` model with a negative amplitude value. This negative value is handled internally, so what the user sees in the GUI and what is reported in the yaml model file is always a positive value.

The main difference is in the initialization, where `GaussianAbsorption` depends on an internally defined "continuum".

Issue #275, opened 2017-03-30
BUG: reading multiple files by emitting 'on_file_read'


astrofrog:
I think there is an issue when multiple ``on_file_read`` events are fired in quick successions. The following example illustrates the problem:

```python
import glob

from qtpy.QtWidgets import QApplication

from specviz.core import dispatch
from specviz.ui.viewer import Viewer

app = QApplication([''])

viewer = Viewer()

for filename in glob.glob('*spec*.fits'):

    print('Loading spectrum from ', filename)

    dispatch.on_file_read.emit(file_name=filename,
                               file_filter='MOS')

viewer.main_window.show()
app.exec_()
```

Note that because of #274 I had to make a directory with just the spectra and run the script located elsewhere, by doing e.g. ``python ../script.py`` (otherwise there was some kind of recursion). When running this, only one (the last) spectrum is read into specviz. If I add a ``time.sleep(1)`` inside the loop, then multiple datasets show up.

This causes the glue plugin to not work correctly and only show one of the spectra from the selected subsets.

nmearl:
This was expected (but not ideal). It's a consequence of keeping access to a single thread. If the thread doesn't return before it's called again, it will ignore the initial inputs.

There is a quick fix for this.

nmearl:
@astrofrog Try your script with the latest merge.

astrofrog:
It works! I have a new related issue though, but I think it's a separate bug - will open a new ticket.

Issue #274, opened 2017-03-30
BUG: specviz tries to run all python scripts in current working directory


astrofrog:
If I install the latest version of specviz and try and run ``specviz``, it will try and run every single Python script in the current working directory. For example if I put a print statement at the top of scripts called 'test.py' and 'script.py' in the current working directory and start specviz, I get:

```
$ specviz
INFO:root:Added Simple Fits to loader registry.
INFO:root:Added HST/COS to loader registry.
INFO:root:Added HST/STIS to loader registry.
INFO:root:Dispatch is now watching: <specviz.ui.widgets.windows.MainWindow object at 0x1209eb048>
RUNNING SCRIPT
INFO:root:Dispatch is now watching: <specviz.ui.widgets.windows.MainWindow object at 0x12c944828>
WARNING:root:Unable to import race_condition in /Users/tom/Data/Spectra.
running script.py
INFO:root:Dispatch is now watching: <specviz.ui.widgets.windows.MainWindow object at 0x12c944ee8>
WARNING:root:Unable to import script in /Users/tom/Data/Spectra.
running test.py
```

Note the 'running' messages. I don't think this would be expected behavior by users and should IMHO be fixed for the release. If you want specviz to automatically load certain kinds of models, I would recommend an approach like in glue where we whitelist specifically a certain filename (``config.py``). Maybe here you could whitelist ``specviz_models.py`` or similar?

nmearl:
This is somewhat intentional, see #185. I'm not sure whitelisting is practical since there is no imposed naming convention for python custom loader files.

The middle ground could be to load *only* yaml files from the current directory.

Personally, I think the user should be responsible for moving specviz-specific loader files to their `~/.specviz` directories.

@hcferguson @pllim Thoughts?

astrofrog:
I think loading all Python scripts is definitely going to cause a lot of confusion, and I would support only loading Python files from a ``.specviz`` directory. Note that for custom loader files, you could certainly choose to impose a convention, for example that the name has to start with specviz?

mattjhill:
I would also vote for only loading files from the `.specviz` directory

hcferguson:
I think loading all python scripts creates a gaping security hole. 
Are the YAML loader files executable? I thought they were just being read in and we had disabled the ability to include executable code in them.

hcferguson:
That said...I think there is a lot of value in being able to have the custom loaders in the same directory as the data. So what is the downside of making the YAML files not executable?

[...]

PR #277, opened 2017-03-30
File loading and loader parsing bug fixes

nmearl:
Addresses issues described in #275 and #274.


nmearl:
@mattjhill When you get a change, could you review this please?

PR #279, opened 2017-03-30
Docs

mattjhill:
Updating docs to: 
 - use the current button names
 - include info on fixing model parameters
 - change some wording on landing page
 - use new demo video on Box


mattjhill:
Also put sphinx in environment.yaml to use an up to date version.  Default on RTD is 1.3.5 which doesn't use the same math extension names and crashes during build.

PR #285, opened 2017-04-05
Travis update

nmearl:
Travis is being difficult, using this PR to mess with settings.


nmearl:
@mattjhill When you get a chance?

PR #286, opened 2017-04-21
SDSS loaders

### LINKED ISSUE #261: Custom loader for APOGEE spectra
I'm trying to load APOGEE spectrum files (specifically [aspcapStar files](https://data.sdss.org/datamodel/files/APOGEE_REDUX/APRED_VERS/APSTAR_VERS/ASPCAP_VERS/RESULTS_VERS/LOCATION_ID/aspcapStar.html)) into specviz.  These are formatted very differently from SDSS/BOSS "spec" files.  To summarize:

1. Data and WCS headers are in HDU 1 (HDU 0 is empty), which is a 1-d image, not a table column.
2. The data values are arbitrary units; these spectra are pseudo-continuum normalized.
3. Wavelength solution in WCS is Log-linear.
4. Uncertainties are in HDU 2, which is also a 1-d image.
5. The best-fit spectrum is in HDU 3. I'd like to be able to extract that along with the data spectrum.

Here's what I have so far:

```yaml
--- !CustomLoader
name: APOGEE aspcapStar
extension: fits
wcs:
    hdu: 1
dispersion:
#     hdu: 1
#     col: 1
    unit: 'dex(Angstrom)'
data:
    hdu: 1
#     col: 0
#     unit: '1e-17 erg / (Angstrom cm2 s)'
uncertainty:
    hdu: 2
#     col: 2
#     type: 'ivar'
meta:
    author: Benjamin Alan Weaver
```

With this loader, I get an error: `ERROR:root:Incompatible loader for selected data: APOGEE aspcapStar because 'col'`

weaverba137:
This PR:

* Closes #261 
* Slightly modifies the existing SDSS optical spectrum YAML loader to note that it only applies to SDSS-III/IV spec files.  SDSS-I/II spSpec files should not use that loader.
* Adds support for several types of APOGEE spectra.

While working on this, I had a question: is it possible to independently scale the x and y (wavelength & flux) axes?  Although the spectra appear to be loaded and displayed "correctly", they look terrible, because the scale of the *errors* appears to set the scale of the y axis, and the errors can sometimes be rather large.  In other words, I'd like to zoom the flux axis while not zooming the wavelength axis.


weaverba137:
The documentation error in the Travis test does not appear to be related to this code.  What are the current plans to fix this?

specviz/io/loaders/apogee_loader.py
```diff
@@ -0,0 +1,174 @@
...
+from astropy.io import fits
+from astropy.table import Table
+from astropy.wcs import WCS
+from astropy.units import Unit, def_unit
+from astropy.nddata import StdDevUncertainty
+
+import numpy as np
+
+from specviz.interfaces import data_loader
+from specviz.core.data import Spectrum1DRef
```
weaverba137: Yes, I'll change that.

Issue #200, opened 2016-05-20
Specify custom loader columns by name


sean-lockwood:
Feature request, not a bug...

Rather than specifying a custom loader's column via numerical position, you could use the FITS column name.  Instead of:
```
dispersion:
  hdu: 1
  col: 2
```
It would be nice to specify:
```
dispersion:
  hdu: 1
  col: 'WAVELENGTH'
```
(Additionally, I believe the FITS standard makes column names case-insensitive.)

This is helpful because a loader would apply to more file types and require less code replication, assuming column names are the same.

pllim:
Yeah, that is a reasonable request. Personally, not sure if this YAML thing is the way to go. General feedback we get is that it is hard to use. Perhaps we don't need this anymore if we switch to `specutils` in the future...

nmearl:
> Perhaps we don't need this anymore if we switch to specutils in the future...

I'm not sure how `specutils` would offer anything remotely similar to custom loaders?

nmearl:
The alternative would be to implement custom loaders in python, and have the user be responsible for constructing the `Data` object that then gets stored.

pllim:
We might have to rethink this for long term support. This is a GUI application, not a universal table loader. The loader stuff should be somewhere else if it is reasonably complicated. At some point, we just have to draw the line and say something like, "If <third party package> cannot read this, you have to reformat your crazy data."

pllim:
Anyways... tl;dr -- Good request *if* we decide to stick with custom YAML for every format out there.

ibusko:
This issue is 2 years old. There seems to be now a good overlap in between the request / suggestion in here, and the spectrum ingestion wizard that was implemented later. I wonder if this could be just closed now.

PR #307, opened 2017-06-15
Support giving column names as strings in YAML

astrofrog:
With this, it's possible to then specify both HDU and columns by name instead of by index, which can be more intuitive:

```yaml
--- !CustomLoader
name: My Reader
extension:
- fits
wcs:
  hdu: PRIMARY
data:
  col: data
  hdu: SPECTRUM
  unit: Jy
uncertainty:
  col: uncertainty
  hdu: SPECTRUM
  type: std
meta:
  author: Tom
```

This also fixes an incorrect logging message that included the variable 'filter' which was no longer passed.


nmearl:
Related to #200.

astrofrog:
The failure is related to specutils being out of date with astropy master (which no longer includes OrderedDict) - I think we'll need to figure out a fix before the specviz release.

PR #308, opened 2017-06-15
Import Wizard

astrofrog:
**Note: this is a work in progress and should not be merged**

This is an import wizard that can make it easy for users to determine how to read their files. Currently it will simply print out the required YAML to the terminal, but in future of course the data could then load automatically into specviz.

screen shot 2017-06-15 at 15 06 47

Things that need to be done:

* [x] Make it possible to launch wizard from Specviz UI
* [x] Make it possible to save the yaml configuration to ``.specviz``
* [x] Make it possible to disable uncertainties
* [x] Automatically read spectrum into specviz once user clicks 'ok'
* [ ] Remove units from uncertainty section (?)
* [ ] Make it possible to load in mask


specviz/io/wizard/wizard.py
```diff
@@ -0,0 +1,483 @@
...
+    dispatch.on_file_read.emit(file_name=filename,
+                               file_filter=yaml_filter)
+
+    # Remove YAML loader - this is a big hack at the moment since there is no
+    # official way in Astropy to unregister a reader/writer/identifier
+    # FIXME: how do we determine when the file has been read?
+    import time
+    time.sleep(1)
+    io_registry._readers.pop((yaml_filter, Spectrum1DRef))
+    io_registry._identifiers.pop((yaml_filter, Spectrum1DRef))
```
astrofrog: @nmearl - I'm temporarily registering a YAML custom reader and then want to remove it after reading, but since things happen in threads I'm not sure what the best way is to know once the data has been read and the reader can be removed. Do you have any recommendations?

astrofrog: Also I just opened https://github.com/astropy/astropy/pull/6217 to fix the ability to unregister readers

nmearl: There a couple ways off the top of my head. One would be to listen for `on_added_data` events, since those are called after a data set has been loaded from a file:

```python
dispatch.on_file_read.emit(file_name=filename, file_filter=yaml_filter)

def remove_yaml_filter(data):
    if data.name != filename:
        return

    io_registry._readers.pop((yaml_filter, Spectrum1DRef))
    io_registry._identifiers.pop((yaml_filter, Spectrum1DRef))

    dipatch.unregister_listener("on_added_data", remove_yaml_filter)

dispatch.register_listener("on_added_data",  remove_yaml_filter)
```

Alternatively, you could just spin up your own thread:

```python
from specviz.core.threads import FileLoadThread

file_load_thread = FileLoadThread()

# Connect to the status bar in the gui, if you like
file_load_thread.status.connect(
    dispatch.on_status_message.emit)

# Define some result handling function
def file_load_result(data):
    dispatch.on_add_data.emit(data=d)

    io_registry._readers.pop((yaml_filter, Spectrum1DRef))
    io_registry._identifiers.pop((yaml_filter, Spectrum1DRef))

# Hook up thread
file_load_thread.result.connect(file_load_result)

# Call and start thread
file_load_thread(filename, yaml_filter)
file_load_thread.start()
```

astrofrog:
The Travis failure with astropy dev is unrelated, so I'll go ahead and merge this. I'll investigate the astropy dev failure soon, but just ignore it for now.

PR #313, opened 2017-06-22
Add support for selecting mask information in wizard

astrofrog:
Note that the wizard code needs to be tidied up and commented, but I'm just focusing on getting the functionality in there first, and will do a tidy-up in a separate PR.

cc @mattjhill


mattjhill:
Looks good to me.

PR #314, opened 2017-06-22
Fix Astropy dev errors on Travis

astrofrog:
This attempts to fix the failures with Astropy dev, which actually have nothing to do with Astropy dev itself but more the fact that Astropy no longer bundles pytest, and we are therefore using the latest version of pytest.

Note that ``doctest_plus = disabled`` doesn't have the intended effect, which I've reported in https://github.com/astropy/astropy/issues/6268

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ env:
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
         - ASTROPY_VERSION=stable
+        - SETUP_XVFB=true
         - SETUP_CMD='test'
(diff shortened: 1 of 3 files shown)


astrofrog:
@mattjhill @nmearl - this is ready to merge and fixes the tests on Travis

nmearl:
I'd like to merge the changes from my fork before this, since it incorporates some higher-up changes from the astropy affiliate template.

astrofrog:
@nmearl - no problem

astrofrog:
I think this was fixed in the mean time

PR #312, opened 2017-06-22
Beach Release Sprinting

nmearl:
This PR involves changes to the structure of the repository, updates some old Astropy affiliate template files, includes some switch over to Qt Designer-based gui files instead of hand-coded, incorporates the advanced fitter settings and the resampling functionality, and beings the work of a richer command-line interface.

PR #315, opened 2017-06-23
WIP: Mask editor

mattjhill:
Still needs some work, and should wait until after #312.


mattjhill:
@astrofrog Can you remind me the new format you had in mind for the YAML files for the mask table?

astrofrog:
I think we said that in the ``mask`` section there would simply be e.g. ``definition: SDSS`` for example, where definition is just the name of a registered bit mask description. We could then set up a registry for bit mask descriptions where we can include SDSS and JWST by default and allow users to register their own?

mattjhill:
closing in favor of #317

PR #316, opened 2017-06-24
Fix loading of HDUs with no name

astrofrog:
This fixes the issue with ``generic_spectra.fits`` which was to do with the fact that HDU 1 had no name (which is fine).

@nmearl @kassin - the ``generic_spectra.fits`` file also contains pseudo-WCS in HDU 0, but this is incomplete:

```
CUNIT1  = 'Angstrom'                                                            
CRVAL1  =                200.0                                                  
CDELT1  =                  1.0                                                  
```

specifically, CTYPE1 is missing, so astropy WCS doesn't recognize this as being a spectral WCS. Is this a fake spectrum file, or was it found in the wild?

### DIFF
--- a/specviz/utils/parse_fits.py
+++ b/specviz/utils/parse_fits.py
@@ -24,7 +24,7 @@ def parse_fits(filename):
 
     with fits.open(filename) as hdulist:
 
-        for hdu in hdulist:
+        for ihdu, hdu in enumerate(hdulist):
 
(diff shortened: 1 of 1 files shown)


nmearl:
I completely made this up, so I'm not surprised that there are missing header keywords. I used only enough that I could compose a dispersion array internally.

PR #344, opened 2017-11-30
scipy not used for spectral statistics

migueldvb:


brechmos-stsci:
This file appears to be re-factored. So this PR isn't directly relevant. If there are other similar code things, we will have to find as part of a separate cleanup.

PR #359, opened 2018-02-12
Fix to show position update

brechmos-stsci:
The 'Pos:' was not being updated on the bottom toolbar when the mouse is moved around the screen. It appears to have just been a conditional check as the `index` value is == 0 in the normal case.  This fix works in the third_party glue attachment in cubeviz.

image


nmearl:
Interesting, I'll look into this further another time -- the conditional doesn't affect the position text in specviz proper.

PR #372, opened 2018-03-15
Show/hide the median data based on existence of other subests

brechmos-stsci:
Hide the median data cube spectrum if there are other subsets.  Then show the median data cube spectrum if there are no other subsets.


brechmos-stsci:
The one thing not in here is autoscaling of the y-axis which might be really nice to have as it will be something people will want to have (automatically). Going to look into that now.

brechmos-stsci:
@nmearl I would suggest let's merge this and then I'll look at the auto-scaling later.

PR #373, opened 2018-03-16
Thicker plot lines and thicker slider line

brechmos-stsci:
There were several requests to make the plot lines thicker as they are hard to see in the cubeviz window. I think it might make sense here if you are good with it. I also made the green line, that represents the spatial slice location, thicker as it was hard to see.

### DIFF
--- a/specviz/core/plots.py
+++ b/specviz/core/plots.py
@@ -51,7 +51,7 @@ class LinePlot(object):
         self._plot_units = (self._layer.dispersion_unit,
                             self._layer.unit,
                             None)
-        self.line_width = 1
+        self.line_width = 3
         self.mode = None
         self.checked = True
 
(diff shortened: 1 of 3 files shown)


nmearl:
Plot lines can already be made thicker using plot settings in SpecViz.

brechmos-stsci:
Oh... didn't see that, can you post an example?  I don't see it. Is it in `SpecVizViewer`?

nmearl:
screen shot 2018-03-16 at 3 04 50 pm

brechmos-stsci:
Oh, yeah, knew that was there. I am looking for a programmatic one so the default (in cubeviz) is a thicker line.

nmearl:
You should not be doing that here, then, since this would affect specviz as a whole. While it's fine to change the wavelength slider width in the code, to change the default plotted line width, you should add it into the style parameters on line 240 (and pull the latest specviz):

```python
dispatch.on_add_to_window.emit(data=spec_data,
                               style={'color': layer.style.rgba[:3],
                                      'line_width': 3})
```

brechmos-stsci:
Ok, so I changed it so that it is only the slider line being thicker.  I'll look at changing the other ones via cubeviz.

PR #377, opened 2018-03-19
Pass message back out if there is a change

brechmos-stsci:
I needed to receive a message in cubeviz that there was a change in redshift Z.  (If there is a better way to do this, please post, I'll fix.)

This works for the cubeviz stuff, not sure if it will break other specviz stuff.


brechmos-stsci:
I modified the comment.

Issue #376, opened 2018-03-19
Spectral smoothing Standard Deviation needs units


brechmos-stsci:
The users are not finding it clear what the units are for the Standard Deviation (pixels vs actual astronomical units).

![image](https://user-images.githubusercontent.com/18348847/37596292-13ea4910-2b52-11e8-9edc-185009e0bf3f.png)

Related to cubeviz view.

ibusko:
Fixed to look like this:

![untitled](https://user-images.githubusercontent.com/3577145/38507579-ed3f3290-3bea-11e8-8102-2904dcd91105.jpg)

Issue #347, opened 2017-12-13
default line weight should be heavier than option 3


kassin:
When spectra are displayed the default line width should be heavier than even option 3.

ibusko:
Made the default 3, and added a '4' option to the menu. '3' looked to me too heavy already, but it could be dependent on the actual data I used for testing.

ibusko:
Reverted for now:

The default line width should stay 1 for performance reasons. Perhaps in the future the default will be settable by the user, but I think we should err on the side of having fast, interactive plots initially.

Issue #333, opened 2017-11-03
Extend the wizard to text files without headers?


hcferguson:
I think it's fairly straightforward to tell if its a fits file or text file, it could do a quick check and then basically pop up a different wizard that lets the user say which column is which and what the units are.

ibusko:
Wizard updated to accept ASCII and ECSV tables.

nmearl:
Closed with #395 and #396.

Issue #331, opened 2017-09-27
Wizard for tabular 1D spectra


kassin:
It would be great to have a wizard that helps readers read in spectra in ascii format
like those in:

/astro/3/jwst_da_sprint_testdata/SpecViz/PN_Muzerolle

astrofrog:
I agree - I wrote the wizard in a way that in principle it could be extended for non-FITS files too, so hopefully it shouldn't be too much work

ibusko:
I bet this could be done easily with the ECSV format.

ibusko:
This could be alternatively done using the ECSV format.

ibusko:
I updated the wizard to read ECSV text files.

ibusko:
Wizard updated to accept ASCII tables.

PR #395, opened 2018-04-12
Small fixes, ECSV spectrum loader and wizard

ibusko:
Implements fixes for #376 and #347.

Adds a bare-bones ECSV spectrum loader and wizard support for ECSV as a first step in implementing #331 and #333.


ibusko:
OK, I reverted the change in the default plot line thickness.

nmearl:
Looks good, please merge when you're ready.

PR #396, opened 2018-04-17
Augment wizard to support ASCII and ECSV tables.

ibusko:
#333 and #331


nmearl:
Does this affect #333?

ibusko:
Yes, it is related to #333. My intent was to provide the ability to use the wizard when no header or column information exists in the file.

PR #428, opened 2018-07-16
Basics for new model-based core; improves on sub window plots; new signaling set up

nmearl:
Introduces new refactor to the master specviz repo.


nmearl:
@javerbukh This is probably too much to review, but since it's a complete refactor of the original, it may be best to accept it as the base, and ensure that all future, smaller, PRs get fully reviewed.

Issue #431, opened 2018-08-13
Creating the plugin for Glue interaction


javerbukh:
Make sure Specviz can be implemented as a DataViewer in Glue

astrofrog:
As mentioned on the call yesterday there are two main aspects to this work:

* Making it so that Specviz can be used as a data viewer for individual spectra loaded into glue

* Making it so that Specviz can be used to visualize spectra with filenames given in the column of a data file loaded into glue (the previous behavior)

astrofrog:
Actually only the first should work here, but also with support for spectral cubes - this is done in #436. The second item is specific to MOSViz.

PR #436, opened 2018-08-15
Add back glue plugin functionality

astrofrog:
This re-implements the glue plugin, with some differences:

* It's now possible to visualize glue data of any dimensionality as long as it has a spectral axis (rather than just a cube)
* The modeling panel hasn't been added back yet (it might be easier to do this in a separate PR)

To use, open a spectral cube or a single spectrum in glue and create a Specviz viewer with it:

screen shot 2018-08-19 at 15 58 14

Remaining to-dos:

* [x] Rebase once #445 and #446 are merged
* [x] Fix issue of plot refreshing when set_data is called

Fixes https://github.com/spacetelescope/specviz/issues/431


specviz/third_party/glue/viewer_single_spectrum.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # FIXME: simplify the following by implementing DataListModel.remove_data
+        # and making it able to take e.g. data name.
+
+        data_model = self.specviz_window.workspace._model
+        for i in range(data_model.rowCount()):
+            item = data_model.item(i)
+            if item.data() == self.uuid:
+                data_model.removeRow(i)
+                return
```
astrofrog: @nmearl - this is where it would be handy to be able to remove a data object from Specviz based on its name (which here is a a UUID)

specviz/third_party/glue/viewer_single_spectrum.py
```diff
@@ -0,0 +1,170 @@
...
+        Get the PlotDataItem corresponding to this layer artist.
+        """
+
+        # FIXME: this definitely needs to be simplified!
+
+        proxy_index = self.proxy_index
+        if proxy_index is None:
+            return
+        else:
+            return self.specviz_window.workspace.proxy_model.item_from_index(proxy_index)
```
astrofrog: @nmearl - here it would be nice to have a way to get a PlotDataItem given the dataset name (which here is a UUID)

specviz/third_party/glue/viewer_single_spectrum.py
```diff
@@ -0,0 +1,170 @@
...
+    def update(self, *args, **kwargs):
+
+        if not is_glue_data_1d_spectrum(self.state.layer):
+            self.disable('Not a 1D spectrum')
+            return
+
+        self.enable()
+
+        if self.plot_data_item is not None:
+            self.clear()
```
astrofrog: @nmearl - currently I'm removing and adding data back again if the data values are updated. Is there a way to change the data in-place? Can I just change the data in the Spectrum1D object in-place?

specviz/third_party/glue/viewer_single_spectrum.py
nmearl: Would you prefer something like

```
self.specviz_window.workspace.plot_data_item_from_id(id)
```

or

```
self.specviz_window.workspace.proxy_model.item_from_id(id)
# And relatedly, for DataItems
self.specviz_window.workspace.model.item_from_id(id)
```

specviz/third_party/glue/viewer_single_spectrum.py
nmearl: So you're looking for the ability to update the `Spectrum1D` stored in the `DataItem`?

specviz/third_party/glue/viewer_single_spectrum.py
nmearl: I've implemented the method `data_model.remove_data(<id>)` in #441.

specviz/third_party/glue/viewer_single_spectrum.py
nmearl: If, I've added a method to the `DataItem` object:

```python
data_item.set_data(new_spectrum1D)
```

astrofrog:
@nmearl - thanks for the programmatic API improvements! Things are a lot simpler here now. The only remaining issue for now is that when calling ``set_data`` to update the data, the plot doesn't update straight away - I need to set ``visible`` on the ``PlotDataItem``s to ``False`` then ``True`` again for the changes to take effect. Any idea what is causing this?

PR #438, opened 2018-08-15
Adding arithmetic widget and ui

mfixstsci:
PR for the arithmetic widget. It's not perfect but the different views are able to communicate to one another. Still needs to be able to validate/evaluate user input.


mfixstsci:
Going to submit another pull request with arithmetic widget. This request still relies on glue and still needs some overhauling. Also, git branches were crossed and tangled up.

PR #440, opened 2018-08-16
Smoothing UI

### LINKED ISSUE #411: Re-implement spectral smoothing
- Signal/slot creations
- Threading

robelgeda:
Closes #411:
This PR introduces smoothing functionality and a smoothing UI to SpecViz. Smoothing operations are performed on a `QThread` via the smoothing functions in `specutils.manipulation.smoothing`. The smoothing option can be found under the operations menu. Output are added to the workspace after smoothing and are labeled: 
`<original name> Smoothed(<kernel type>, <kernel size>)`

**Smoothing UI:** 
screen shot 2018-08-16 at 10 11 38 am

**Available Kernels:**
- Box
- Gaussian
- Trapezoid
- Median


nmearl:
One thing I noticed when you were demoing is that the dialog gives a drop-down of all the available data sets. Currently, smaller things like the line color change relies on the _currently highlighted_ item in the data list (so a user would select the data item they want to perform the operation on, and then click the button to perform that operation).

I'd like to stick to one or the other -- either have the operation be agnostic to the selected data in the list (like how you have it here) and provide a drop down; or have the user select the data item first, and then the operation they wish to perform on it (this is how e.g. things like Photoshop work).

Thoughts?

robelgeda:
I think we can blend the two ideas this way: when the dialog pops up, the drop down is set to the current selected data but the user can still change the data if want. That way, the user doest have to relaunch the smoothing dialog incase they miss-selected. I think the drop down is helpful because you know what your smoothing before you hit the button. Would this work?

nmearl:
Definitely, I think that's a good middle-ground.

robelgeda:
Alright it should be ready now @nmearl

nmearl:
@robelgeda Awesome, I did a PR against your fork that incorporates the simpler menu creation.

robelgeda:
Ok, I merged your PR.

PR #445, opened 2018-08-18
Add the ability to customize the line width and zorder

astrofrog:
The line width works but not the zorder - @nmearl, do you have any ideas why that might be?

### DIFF
--- a/specviz/core/items.py
+++ b/specviz/core/items.py
@@ -61,6 +61,7 @@ class PlotDataItem(pg.PlotDataItem):
     data_unit_changed = Signal(str)
     spectral_axis_unit_changed = Signal(str)
     color_changed = Signal(str)
+    width_changed = Signal(int)
     visibility_changed = Signal(bool)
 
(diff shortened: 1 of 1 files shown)


nmearl:
Hmm, not sure. I went ahead and tried locally by just adding

```python


    @property
    def z_value(self):
        return self.zValue()

    @z_value.setter
    def z_value(self, value):
        self.setZValue(value)
```

to the `PlotDataItem`, and adding a plot action via

```python


        z_act = QAction("Z Value", self._central_widget)
        self._central_widget.tool_bar.addAction(z_act)

        def inc_z():
            self.current_item.z_value = self.current_item.z_value + 1

        z_act.triggered.connect(inc_z)
```

to `PlotWindow`, and I can control the z order of the plots.

I'll continue to investigate.

astrofrog:
@nmearl - actually I think it does work, it's just that I'm finding I have to hide and show the layers again for things to take effect.

PR #429, opened 2018-08-10
Refactor unit conversion

javerbukh:
This PR has the UI and basic functionality for unit conversion, but I still have questions as to using Signals for data conversion and where the data will be stored.

PR #448, opened 2018-08-23
Specviz gui tests

javerbukh:
specviz_gui is created in conftest so that we can write tests that interact with the gui. Just use specviz_gui as an argument in a test method and you will have access to the gui.


specviz/data/ui/plot_window.ui
```diff
@@ -71,6 +71,7 @@
    <addaction name="remove_region_action"/>
    <addaction name="plot_options_action"/>
    <addaction name="export_plot_action"/>
+   <addaction name="change_unit_action"/>
```
nmearl: Are these unit dialog ui changes different from the ones in your other PR?

specviz/widgets/plotting.py
```diff
@@ -1,20 +1,25 @@
...
-                            QToolButton, QSizePolicy, QWidget, QMenu, QWidgetAction, QColorDialog, QMessageBox)
+                            QToolButton, QSizePolicy, QWidget, QMenu, QWidgetAction, QColorDialog, QMessageBox, QDialogButtonBox)
 from qtpy.uic import loadUi
 
 from ..core.items import PlotDataItem
 from ..core.models import PlotProxyModel
 from ..utils import UI_PATH
 from .custom import LinearRegionItem
 
+logging.basicConfig(level=logging.DEBUG, format="%(filename)s: %(levelname)8s %(message)s")
```
nmearl: Are these different than those in #429?

specviz/data/ui/plot_window.ui
javerbukh: Nope

specviz/widgets/plotting.py
javerbukh: No, I'm removing them now

javerbukh: I've noticed you used logging in the setter functions for data_unit and spectral_axis, should I just leave it in?

nmearl:
Could you also remove the `.pytest_cache` files here too?

javerbukh:
@nmearl They are removed

javerbukh:
I'm not sure how, I haven't merged or made any changes to this since I reverted the helpers last time

javerbukh:
Made outdated by PR #489

PR #450, opened 2018-08-27
Adding arithmetic layer

mfixstsci:
Still needs to be hooked into SpecViz via the arithmetic button. Right now, `arithmetic_editor.py` can be run from the command line and will create 2 `DataItem` objects which are placed in a list to simulate a `DataModel` object that gets passed to the arithmetic layer. What needs to be done:

- Load arithmetic layer UI when arithmetic button is clicked

- Make sure DataModels loaded in SpecViz session are passed to arithmetic layer when launched.

- When a calculation is made and arithmetic layer exits, how do we store the values calculated in SpecViz?

- Docstrings need work/updating


specviz/plugins/arithmetic/__init__.py
```diff
@@ -0,0 +1 @@
+from .arithmetic_editor import *
```
robelgeda: New line needed

specviz/widgets/arithmetic_editor.py
```diff
@@ -0,0 +1,230 @@
+import astropy.units as u
```
robelgeda: This looks like a duplicate file?

specviz/data/ui/equation_editor.ui
```diff
@@ -0,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
robelgeda: This looks like a duplicate? There is another one: `specviz/plugins/arithmetic/equation_editor.ui`

specviz/data/qt/ui/equation_editor.ui
```diff
@@ -0,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
robelgeda: This looks like a duplicate? There is another one: specviz/plugins/arithmetic/equation_editor.ui and specviz/data/ui/equation_editor.ui

specviz/plugins/arithmetic/arithmetic_editor.py
```diff
@@ -0,0 +1,230 @@
...
+    
+        self._equation_editor = equation_editor
+
+        if not self._equation_editor.model_items:
+            self.msgbox = QMessageBox.warning(self._equation_editor, "No Spectrum1D Objects", 
+                                "There is no data loaded into your SpecViz session!")
+        else:
+            self.is_addmode = label is None
+
+            # example=self._equation_editor.model_items
```
robelgeda: Do we need these line?

specviz/plugins/arithmetic/arithmetic_editor.py
```diff
@@ -0,0 +1,230 @@
...
+                self.text_label.setDisabled(True)
+            else:
+                self.text_label.setPlaceholderText("New attribute name")
+            if equation is not None:
+                self.expression.insertPlainText(equation)
+            else:
+                self.expression.setPlaceholderText(self.placeholder_text)
+
+            # Set options for spec1d objects to select.        
+            self.combosel_data.addItems([item.name for item in self._equation_editor.model_items])
```
robelgeda: This should be something like `self.combosel_data.setModel(self._equation_editor.hub.model)` ... That way the combo box is updated when the list of data items changes.

specviz/widgets/arithmetic_editor.py
mfixstsci: Yeah, this one needs to go. The arithmetic editor needs to live in the `plugin` directory.

specviz/data/qt/ui/equation_editor.ui
mfixstsci: Yeah, this needs to go! Was useful here when doing early dev work!

specviz/plugins/arithmetic/arithmetic_editor.py
mfixstsci: Good point, I like this one way more.

nluetzge:
So I had a look at it and it seems to work, at least the examples that are given in the window. Here are a a comments:

1) There is a typo in one of the examples. You are subtracting the spectrum in both examples instead of adding in the first case:

box

2) It took me a while to understand why I get an error when I used the insert button to add the spectrum in the arithmetics box (shown here:)

screen shot 2018-11-05 at 12 42 17 pm

The reason is as it states below that it needs the result to be of the form of a spectrum. Maybe we could add another option to the drop-down menu that just inserts the {COS_FUV} (instead of {COS_FUV}.flux) So that I don't have to write all the time...Maybe we just call it spectrum. 

3) And the last point. When I create an arithmetic and then I'm editing it, it creates a new set of data instead of replacing it. Is this a bug or a feature? Because then I suddenly have two datasets in the left panel that are named exactly the same and that might be confusing...

4) It would be good to test it how well it works with the unit conversion so we might want to rebase it to test this further.

specviz/plugins/arithmetic/arithmetic_editor.py
mfixstsci: I went ahead and deleted these and replaced them with the actual example I wanted to use.

specviz/data/ui/equation_editor.ui
mfixstsci: Another duplicate, it's gone now.

specviz/plugins/arithmetic/__init__.py
mfixstsci: Good catch. I went ahead and updated it to only import the two classes that make up the layer arithmetic!

mfixstsci:
@nluetzge thank you for your feedback!

1. Good catch! Fixed now.

2. I completely agree with you. I was thinking that we could make row just like the `Available Attributes` row to insert spectra. 

3. This is a feature. After speaking with @nmearl he suggested that its probably best that we don't delete or remove data from the collection via the editor and leave that up to the user.

4. I agree, I will make sure to incorporate the upstream changes in the next commit!

PR #455, opened 2018-10-09
Update MANIFEST.in and setup.py to install tracked files

sosey:
I had a problem installing from the git master because it doesn't install the UI files.  This PR updates setup.py and MANIFEST.in to fix the issue. Check the extra lines at the bottom of the MANIFEST.in file to make sure you really want them - I went ahead and ran `check-manifest` just to see what was left out.

Minimally, `recursive-include specviz *.ui` and adding `install_package_data=True` to the setup are needed, but I'm not sure about the others. 

If this PR is in error, feel free to close!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,7 +6,7 @@ include ah_bootstrap.py
 include setup.cfg
 include specviz/tests/coveragerc
 
-recursive-include specviz *.pyx *.c *.pxd
+recursive-include specviz *.pyx *.c *.pxd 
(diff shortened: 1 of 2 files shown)


sosey:
cc @hcferguson

nmearl:
Thanks, this is useful for files that don't exist in the `data` directory (as those includes are handled in the `setup.py` file already [see here](https://github.com/spacetelescope/specviz/blob/master/setup.py#L101)).

PR #459, opened 2018-10-18
Implement old loaders into new specviz

nmearl:


stscicrawford:
Is there a plan to move these into `Specutuils`?

nmearl:
@stscicrawford If that's desired, we certainly can. Perhaps as part of the next specutils release, we can migrate, but we certainly need these in specviz now-ish if for nothing else than testing.

stscicrawford:
That sounds like a good plan and I think these are generally useful to considering migrating upstream when there is time

PR #465, opened 2018-10-23
Fixed TabbedView error

brechmos-stsci:
This fixes the error `AttributeError: type object 'ViewMode' has no attribute 'TabbedView'` that was blocking work. I *believe* this is the correct fix, but it works for now and can be confirmed later.


brechmos-stsci:
It is not clear to me why the glue tests are failing for this. Given that *not* having this PR is going to block work, I suggest it be merged and we figure out the Glue tests next week.

PR #469, opened 2018-10-23
Add a basic test of specviz startup

drdavella:
This is simply intended to catch simple errors like the one that was fixed in #465. It starts the specviz application and that's about it.

Currently there are unrelated test failures that will need to be fixed in order for this test to be meaningful.


javerbukh:
Should we wait until those tests failures you mentioned are fixed?

drdavella:
The current failure comes from the test against Astropy LTS and is fairly arcane (I think I understand it, but continue reading).

Do we really intend to support older versions of Astropy? Or can we just say that we require `astropy>=3.0` for the specviz reboot?

Issue #467, opened 2018-10-23
Specviz Crashes when trying to set a ROI


nluetzge:
ROI in generel seems to be buggy. All it does when clicking on it is creating one big ROI for the entire spectrum. And sometimes when I click on it it crashes the entire program with the error message:

(specviz) Larry:~ nluetzge$ specviz
Traceback (most recent call last):
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specviz-0.6.dev1513-py3.6.egg/specviz/plugins/statistics/statistics_widget.py", line 235, in update_statistics
    spectral_region = clip_region(spec, spectral_region)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specviz-0.6.dev1513-py3.6.egg/specviz/plugins/statistics/statistics_widget.py", line 45, in clip_region
    region.lower = spectrum.spectral_axis.min()
AttributeError: can't set attribute
Abort trap: 6

nluetzge:
@gderosa2004 sees the same things

brechmos-stsci:
I was looking at that error a little more, and that one, at least, I think is due to specviz relying on an old (not released) version of `SpectralRegion` in `specutils`.  I'll take a quick look at this one.

gderosa2004:
Same issue here, can't get past the definition of ROI (gets the entire spectrum and crashes when trying anything on it). 

Traceback (most recent call last):
  File "/Users/gderosa/anaconda/envs/astroconda_3.5/lib/python3.5/site-packages/specviz-0.6.dev0-py3.5.egg/specviz/plugins/statistics/statistics_widget.py", line 235, in update_statistics
    spectral_region = clip_region(spec, spectral_region)
  File "/Users/gderosa/anaconda/envs/astroconda_3.5/lib/python3.5/site-packages/specviz-0.6.dev0-py3.5.egg/specviz/plugins/statistics/statistics_widget.py", line 45, in clip_region
    region.lower = spectrum.spectral_axis.min()
AttributeError: can't set attribute
Abort trap: 6

brechmos-stsci:
We had a screen sharing session and have it understood now. If either end of the spectral region falls outside the extent of the spectrum then it will crash.

PR #473, opened 2018-10-23
Fix for SpectralRegion stats calculation

brechmos-stsci:
There was an error in the code when updating the lower or upper bound of a SpectralRegion. This is not possible in the new code.  The code was re-factored to create a new SpectralRegion object instead. Works locally.

Fixes https://github.com/spacetelescope/specviz/issues/467


brechmos-stsci:
To test this locally, load a dataset, create a region, make the region smaller so it fits on the spectrum. click on the "combined..." in the top left corner and then move the region around to see the stats change.  When the region is moved outside the extent of the spectrum it will show no stats rather than crash.

Issue #452, opened 2018-09-17
UnitChange dialog doesn't handle no units case


nmearl:
The unit change dialog will crash if there are no unit equivalencies. This might be the case if no units were parsed by the data (and therefore the plots units are `None`).

javerbukh:
It seems the functionality is pretty much non-existent with current SpecViz. When trying to load HST/COS data `combined_13330_G130M_v40_bin4.fits` I get the error:
```ERROR:root:'CompositeUnit' object has no attribute 'long_names'
Traceback (most recent call last):
  File "/Users/javerbukh/Documents/specviz/specviz/core/plugin.py", line 209, in <lambda>
    action.triggered.connect(lambda: func(plugin, *args, **kwargs))
  File "/Users/javerbukh/Documents/specviz/specviz/plugins/unit_change/unit_change_dialog.py", line 135, in on_action_triggered
    self.exec_()
  File "/Users/javerbukh/Documents/specviz/specviz/plugins/unit_change/unit_change_dialog.py", line 128, in exec_
    self.setup_ui()
  File "/Users/javerbukh/Documents/specviz/specviz/plugins/unit_change/unit_change_dialog.py", line 149, in setup_ui
    index = self._data_unit_equivalencies_titles.index(self.current_data_unit)
ValueError: 'Joule' is not in list
Abort trap: 6```
I will remove references to 'long_names' as this seems to give problems after every SpecViz update

gderosa2004:
Me and @nluetzge have played a bit with the unit change.
Here is what we have found when using example_generic.fits:

- Conversion of X is currently applied only to errors (not to flux)
![screen shot 2018-10-24 at 12 01 10 pm](https://user-images.githubusercontent.com/15363065/47446433-3b4a0c00-d789-11e8-85af-941e8305ea70.png)

- After one conversion is applied it is impossible to change units again (either X or Y), even if instance spectrum is deleted
![screen shot 2018-10-24 at 12 01 58 pm](https://user-images.githubusercontent.com/15363065/47446504-63396f80-d789-11e8-857b-1c7b1e5bfb5c.png)

- Specviz crashes if conversion not allowed

javerbukh:
Test once #480 is merged.

PR #480, opened 2018-10-24
Refactor unit conversion

javerbukh:
UC Dialog is now functional, with both buttons being enabled. I have tested using the file `combined_13330_G130M_v40_bin4.fits` and most conversions work, but I will continue to work on it. This will solve issue #452


brechmos-stsci:
I am happy to review this when it is ready for review.

brechmos-stsci:
Did a quick review and sent some questions back to @javerbukh on a couple conversion oddities etc.

specviz/core/items.py
```diff
@@ -142,7 +143,7 @@ def is_data_unit_compatible(self, unit):
         return (self.data_item.flux.unit == "" or
                 unit is not None and
                 self.data_item.flux.unit.is_equivalent(
-                    unit, equivalencies=spectral_density(self.spectral_axis)))
+                    unit, equivalencies=spectral_density(self.spectral_axis*u.Unit(self.spectral_axis_unit))))
```
nmearl:
```suggestion
                    unit, equivalencies=spectral_density(self.data_item.spectral_axis)))
```

Functionally equivalent, but a little cleaner.

specviz/core/items.py
```diff
@@ -225,13 +230,23 @@ def visible(self, value):
         self.visibility_changed.emit(self._visible)
 
     def set_data(self):
+        """
+        Sets the spectral_axis and flux. self.flux is called to convert flux units if they had been changed
+        :return:
```
nmearl: Remove `:return:`.

specviz/core/items.py
```diff
@@ -165,9 +166,13 @@ def reset_units(self):
 
     @property
     def flux(self):
-        return self.data_item.flux.to(self.data_unit or "",
+        """
+        Converts data_item.flux - which consists of the flux axis with units - into the new flux unit
+        :return:
```
nmearl: Remove `:return:`.

specviz/core/items.py
```diff
@@ -165,9 +166,13 @@ def reset_units(self):
...
+        """
+        return self.data_item.flux.to(self.data_unit,
                                       equivalencies=spectral_density(
-                                          self.spectral_axis)).value
+                                          self.spectral_axis*u.Unit(self.spectral_axis_unit))).value
```
nmearl: See above suggestion.

specviz/core/items.py
```diff
@@ -184,7 +189,7 @@ def uncertainty(self):
 
         return uncertainty.to(self.data_unit or "",
                               equivalencies=spectral_density(
-                                  self.spectral_axis)).value
+                                  self.spectral_axis*u.Unit(self.spectral_axis_unit))).value
```
nmearl: See above suggestion.

nmearl:
Also, remove the changes to `astropy_helpers`.

specviz/core/items.py
nmearl: Looking at the CI test, it failed because it stalled out when installing pyqt from pip -- I don't think that's related.

brechmos-stsci:
@javerbukh This is a huge change since I last looked at it and it is looking really good. I have noticed any real *bugs* but there are some things that I think need to be looked at BUT could be put off if this PR *must* go in "now"....

This is all using the command: `specviz -F combined_13330_G130M_v40_bin4.fits -L "HST/COS" `

1.  Jy vs mJy:  When I load up a spectrum and then convert to Jy for the flux units, the left side of the plot  y-axis label is actually showing mJy, even though I asked for Jy. Even when I put in custom units of Jy, it still shows me mJy instead.  I think astronomers will find this a little frustrating but @eteq and @hcferguson can weigh on this.  (for that matter, I think we should see Jy, mJy and uJy in the y-axis unit list in the unit conversion dialog)

2. When I change the x-axis units from Angstroms (default loaded) to ergs, the new list of x-axis units is far fewer than when initially loaded and Angstrom is not in the list. I can do the custom unit thing and type in Angstrom, but not sure why the list changes.

3. When I change the default flux units from erg / (A * cm^2 * s) to Jy, I can't get back to the original units (not listed). I can see someone wanting to go back to the original units.

Comments @eteq @hcferguson?  I think you guys have talked about this stuff before so I might be missing some context.  I think these are important to get done on this PR, but if you guys feel it is more important to get this in and create issues for me, I am fine with that. :-)

But, as I said, this is looking pretty good overall.

hcferguson:
My take:

1. Yes, if I load a spectrum in Jy, I'd expect it to show up in the plot in Jy.  And I would like to have Jy, mJy and uJy in the list.

2 & 3 sound like bugs.

javerbukh:
2 I have fixed and 3 I can fix now. I need @nmearl's input for 1. The current functionality is that the user selects Jy and the plot changes it to mJy, where does this happen?

Side note, I can set the option to include prefixes but this it what it looks like:
screen shot 2018-11-01 at 2 50 22 pm

eteq:
re 1: A question about this - the default beahvior is to load whatever units the spectrum actually has, right? (if it has a unit in the wcs), and if it's missing units the unit is set to whatever the loader says it should be?  That is, no "conversion" of the values happens until the user specifically asks?

javerbukh:
Yes, unit conversion is not called unless the user clicks on Unit Conversion icon in the toolbar.

nmearl:
By default pyqtgraph has auto-SI prefixing enabled. This means that it visual represents the labels along an axis as best it can in the nearest SI prefix. This _does not change the unit value of the data_. Certainly, it can be turned off by setting adding `self.getAxis('left').enableAutoSIPrefix(False)` at line 171 of the `plotting.py` file.

3 sounds like an issue with the way @brechmos-stsci is passing into the string unit formation; `A` is not `Angstrom`. This may be why he cannot convert back.

eteq:
Re: 3: I think it's irrelevant what units the spectrum came *from*: if it successfully converts to `Jy`, it should be possible to convert back to  `erg cm^-2 s^-1 Angstroms^-1`regardless of how it got to `Jy`...

hcferguson:
I didn't know auto-prefixing existed. That sounds excellent.

specviz/plugins/unit_change/unit_change_dialog.py
```diff
@@ -32,6 +31,12 @@ def __init__(self, *args, **kwargs):
         self.ui = loadUi(os.path.abspath(
             os.path.join(os.path.dirname(__file__),
                          ".", "unit_change_dialog.ui")), self)
+        self.spectral_axis_unit_equivalencies = []
```
brechmos-stsci: I think a future PR might want to use `self.spectral_axis_unit_equivalencies as a list of tuples (title, equivalency), rather than two lists.  It is too easy for a future programmer to add to one list and not the other and then spend lots of time looking through the code.  I don't know that this is a blocker though.

specviz/widgets/plotting.py
```diff
@@ -168,7 +168,10 @@ def __init__(self, title=None, model=None, visible=True, *args, **kwargs):
         self._region_text_item = pg.TextItem(color="k")
         self.addItem(self._region_text_item, ignoreBounds=True)
         self._region_text_item.setParentItem(self.getViewBox())
+
+        # Removes ability to automatically change units to match science notation
         self.getAxis('bottom').enableAutoSIPrefix(False)
+        self.getAxis('left').enableAutoSIPrefix(False)
```
brechmos-stsci: Is this a fix for https://github.com/spacetelescope/specviz/issues/522 ?  If so, then probably best to add mJy and uJy into the list units one can convert to for the y-axis units.

specviz/plugins/unit_change/unit_change_dialog.py
javerbukh: Is that why the CI tests are failing? That is my main concern for right now.

brechmos-stsci:
It sounds like with 2 and 3 fixed and 1 will be addressed in https://github.com/spacetelescope/specviz/issues/522 then I think this can be merged once the tests are fixed.

specviz/widgets/plotting.py
javerbukh: It is. Like I said before, I can set `include_prefix_units=True` but the result is the picture I showed before.

javerbukh:
@brechmos-stsci My question remains, *how* will the tests be fixed? It is frustrating to be told my commit is failing tests but the only error message I get is that `No output has been received in the last 10m0s`. What am I supposed to do with that information?

javerbukh:
So the latest commit solves 1, 2, and 3. It also alphabetizes the units in the combobox and even presents Angstrom as such instead of Aa (which is more difficult than it should be). This PR also solves #508 and #452 and would allow @gderosa2004 and @nluetzge to continue testing.

Please review ASAP @brechmos-stsci @nmearl @drdavella

brechmos-stsci:
Ohhh... also under the assumption that test and docs PR will be added very soon.

PR #483, opened 2018-10-24
specify specutils version

### LINKED ISSUE #464: Specify minimum version of specutils in setup.py
I'm assuming specviz doesn't work with older versions of specutils - if so, we should add the minimum version in the ``install_requires`` at https://github.com/spacetelescope/specviz/blob/master/setup.cfg#L51

eteq:
This should fix #464 - this is actually an "aspirational" requirement because I'm not sure it's currently true... but I know some of the current/near-term PRs will require it.


eteq:
At the tagup we just discussed and I think everyone agreed ``>=0.4`` is currently the right answer

PR #489, opened 2018-10-25
Specviz gui tests

javerbukh:
Opens specviz_gui using pytest. Might be redundant with Dan's most recent PR.


specviz/tests/test_specviz_gui.py
```diff
@@ -0,0 +1,9 @@
+import pytest
```
drdavella: This import actually isn't necessary at the moment.

PR #491, opened 2018-10-26
Force users to choose loader before selecting file in dialog

drdavella:
This is intended to prevent accidental use of wrong file type when using the Load Data dialog.

I realize this goes against what we discussed yesterday, but there are no tests here. This is going to be very difficult to test in an automated test suite since the dialog is actually a canned one from Qt itself, and it runs synchronously in the main thread. This means we couldn't even attempt to access this dialog without running a separate thread within our unit tests and yuck.

### DIFF
--- a/specviz/widgets/workspace.py
+++ b/specviz/widgets/workspace.py
@@ -316,13 +316,20 @@ class Workspace(QMainWindow):
         :class:`~specutils.Spectrum1D` object and thereafter adds it to the
         data model.
         """
-        filters = [x['Format'] + " (*)"
+        # This ensures that users actively have to select a file type before
+        # being able to select a file. This should make it harder to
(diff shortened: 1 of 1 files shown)


drdavella:
Also worth mentioning that this is a *slight* hack, but it has the intended effect and I think it's okay. Others may disagree.

nmearl:
@drdavella the way to do this in Qt is by adding the file filter to the name of the displayed loader. E.g. "HST/COST (*.fits, *.fits)". This will auto-filter the displayed filed list and disable any file that does not match this pattern.

The old specviz did this by adding an `extension` keyword argument to the `data_loader` decorator that is used in the custom loaders. However, that doesn't really have any use from the `specutils` side and so hasn't been supported/implemented yet. An option would be to somehow parse the file extension in the `identifier` function used in the astropy registry.

Issue #474, opened 2018-10-23
Stats text areas don't properly show numbers


brechmos-stsci:
The formatting needs to be fixed to properly show the numbers.

![image](https://user-images.githubusercontent.com/18348847/47385213-67a55000-d6d7-11e8-986f-edf358eca58b.png)

javerbukh:
Resolved by #492.

Issue #461, opened 2018-10-22
Discuss what Spectral Analysis Functions will be in October Release


javerbukh:
Issue continues from #425 . Determine what spectral analysis functions need to be in SpecViz this release (October Sprint).

gderosa2004:
I talked with @nluetzge, I think that the basic measurements are critical for a release:

**CRITICAL:**
- S/N
- Center/Centroid
- FWHM 
- EW
- Mean & Std dev in a wavelength region
- Flux 

About the flux (integral of the flux density in a given region): scientists are usually interested in the flux of emission lines that are continuum subtracted. As a zero order analysis we do a linear interpolation (or linear fit) of the continuum in 2 windows close to the emission line, subtract and then integrate the flux of the line directly from the data (no modeling involved). It looks like all the functionalities are already implemented in specutils, just need to re-arrange the workflow.

eteq:
Looking at this, I think all of the "Critical" items are in already - the ones @SaOgaz identified as "not yet in" I think comes from the issue tracker, and it turns out there were some older duplicate issues that we forgot to close after the 0.4 release and merging of the relevant PRs. I *think* I've now fixed that up, but just to be sure, here's a summary of the situation and the relevant parts of the docs: 

* S/N - https://specutils.readthedocs.io/en/latest/analysis.html#snr
* Centroid - https://specutils.readthedocs.io/en/latest/analysis.html#centroid
* FWHM - https://specutils.readthedocs.io/en/latest/analysis.html#line-widths (note that this might needs some further inspection from @gderosa2004 and @nluetzge to make sure you agree the implementation here is sufficient)
* EW - https://specutils.readthedocs.io/en/latest/analysis.html#line-flux-estimates
* Mean & SD in a wavelength region - see the discussion in #243 and #244 where I give one-liners that do this and argue that we may not want them as stand-alone functions because it sort of is a mode of user training. But the point for this is that `specviz` can use my one-liners
* Flux - also in https://specutils.readthedocs.io/en/latest/analysis.html#line-flux-estimates

Note that for flux/eqw the current implementation is not optimal because it doesn't work with fractional pixels - see #365 for more on that, but I'm pretty sure we don't consider that in the "critical" path, agree @gderosa2004 ?

gderosa2004:
Agreed @eteq, fractional pixels are not critical for this release.

[...]

Issue #462, opened 2018-10-22
Get Example Loadable Data for Internal Testing


javerbukh:
ASCII; FITS: COS, STIS, WFC3/GRISM; WCS

gderosa2004:
It looks like there is already a folder with test data collected from a variety of users [https://github.com/spacetelescope/specviz/wiki/Test-Data](url). I think it would be better to restrict the sample for the moment. I have prepared 4 simple spectra that are ready to be updloaded. @javerbukh do you prefer central storage, or box, or anything else?

javerbukh:
Thats a good question, could you reiterate it at the tag-up today or on slack?

gderosa2004:
Test data are contained in here: /astro/3/jwst_da_sprint_testdata/SpecViz/testdata_201810

example_stis.fits 	STIS spectrum, GD71 (Star)
example_cos.fits 	COS spectrum, NGC5548 (AGN)
example.ascii 		UV spectrum, NGC5548 (AGN)
example_generic.fits 	test spectrum, generic fits file

Will increase the sample if needed.

gderosa2004:
Data are also available in box: https://stsci.app.box.com/folder/56430948756

javerbukh:
More test data from Ivo: `/eng/ssb/busko/tmp/combined_13330_G130M_v40_bin4.fits` of type HST/COS

gderosa2004:
@javerbukh @ibusko the combined*.fits spectrum is mine, it's slightly manipulated and not standard calCOS output. I think it would be better to use the standard cos output currently in the testdata folder

gderosa2004:
Test data have been updated both on central storage (/astro/3/jwst_da_sprint_testdata/SpecViz/testdata_201810) and box (https://stsci.box.com/s/zz2vgbreuzhjtel0d5u96r30oofolod7):

STIS_FUV.fits 	STIS FUV/MAMA spectrum, GD71 (Star)
STIS_NUV.fits 	STIS NUV/MAMA spectrum, GD71 (Star)
STIS_CCD.fits 	STIS optical spectrum, GD71 (Star)
COS_FUV.fits 	COS FUV spectrum, NGC5548 (AGN)
COS_NUV.fits 	COS NUV spectrum, GD71 (Star)
example_generic.fits 	test spectrum, generic fits file

SaOgaz:
[...]

PR #492, opened 2018-10-26
WIP: Updating/Adding to analysis functions for ROI sidebar

SaOgaz:
This addresses #461. @robelgeda will be adding the GUI display for the new functions.


brechmos-stsci:
@robelgeda, this is the PR that will fix https://github.com/spacetelescope/specviz/issues/474, right? (or is it going to be a separate PR?

robelgeda:
@brechmos-stsci Yes it will. My part is coming up.

robelgeda:
@SaOgaz : I made a PR at [here](https://github.com/SaOgaz/specviz/pull/1). Once you merge it, it should fix #474 : 
screen shot 2018-10-31 at 12 46 29 pm

SaOgaz:
Handing rest of this PR off to Robel.  Can help with adding tests when we get there. The two other things need that need to be done still are:

- Fixing the updates to the sidebar units when the spectrum display units are changed.
- It looks like we have the stats_widget code repeated in another file: widgets/statistics.py

nluetzge:
I've been playing around with this and I still get a lot of crashes (with the COS_FUV spectrum). Which spectrum are you using to test it? As @gderosa2004  mentioned in #462 we should **not** use the combined spectrum anymore (since this was manipulated and does not reflect the format properly), but the COS_FUV.fits instead.

I'm having trouble to reproduce the crashes. But here are two error messages that I got. It happened when I moved the edges of the ROI or when I moved the ROI.. but not always.

```
  File "/Users/nluetzge/JWST/dev_tools/specviz/specviz/specviz/plugins/statistics/statistics_widget.py", line 318, in update_statistics
    self.stats = compute_stats(spec)
  File "/Users/nluetzge/JWST/dev_tools/specviz/specviz/specviz/plugins/statistics/statistics_widget.py", line 76, in compute_stats
    'fwhm': fwhm(spectrum),
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 105, in fwhm
    return computation_wrapper(_compute_fwhm, spectrum, regions)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/utils.py", line 19, in computation_wrapper
    return func(spectrum, **kwargs)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 173, in _compute_fwhm
    return _compute_single_fwhm(flux, spectral_axis)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 151, in _compute_single_fwhm
    l_idx = np.where(left == True)[0][-1]
IndexError: index -1 is out of bounds for axis 0 with size 0
Abort trap: 6
```

```
  File "/Users/nluetzge/JWST/dev_tools/specviz/specviz/specviz/plugins/statistics/statistics_widget.py", line 318, in update_statistics
    self.stats = compute_stats(spec)
  File "/Users/nluetzge/JWST/dev_tools/specviz/specviz/specviz/plugins/statistics/statistics_widget.py", line 76, in compute_stats
    'fwhm': fwhm(spectrum),
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 105, in fwhm
    return computation_wrapper(_compute_fwhm, spectrum, regions)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/utils.py", line 19, in computation_wrapper
    return func(spectrum, **kwargs)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 173, in _compute_fwhm
    return _compute_single_fwhm(flux, spectral_axis)
  File "/Users/nluetzge/anaconda2/envs/specviz/lib/python3.6/site-packages/specutils/analysis/width.py", line 152, in _compute_single_fwhm
    r_idx = np.where(right == True)[0][0] + argmax
IndexError: index 0 is out of bounds for axis 0 with size 0
Abort trap: 6
```

SaOgaz:
I will investigate crashes....

robelgeda:
@SaOgaz : This should be a ticket in SpecUtils. We can just wrap it around a try-except statement for now.

SaOgaz:
@astrofrog, this travis failure seems to be one of the glue tests. @robelgeda and I tried to figure out why it was failing, something with a copy on one of the WCS objects, but we didn't have much luck.

@hcferguson, is a glue failing test a blocker on merging into master?  (I still need to add tests and get a review before merge)

SaOgaz:
Hey @javerbukh, tests are added, ready for review!

specviz/plugins/statistics/statistics_widget.py
```diff
@@ -57,17 +58,49 @@ def compute_stats(spectrum):
...
-    mean = flux.mean()
-    rms = np.sqrt(flux.dot(flux) / len(flux))
-    return {'mean': mean,
-            'median': np.median(flux),
-            'stddev': flux.std(),
+
+    try:
+        cent = centroid(spectrum, region=None)  # we may want to adjust this for continuum subtraction
+    except Exception as e:
+        print("centroid err: ",e)
```
nmearl: Can you remove this print statement? Also, could you add all the exception output as logging errors?

```python
try:
    ...
except Exception as e:
    logging.error(e)
    ...

File: specviz/plugins/statistics/statistics_widget.py [hunk 2]
```diff
@@ -143,15 +190,17 @@ def _update_stat_widgets(self, stats):
             return
         for key in stats:
             if key in self.stat_widgets:
-                text = format_float_text(stats[key])
-                self.stat_widgets[key].setText(text)
+                print(key, " ", stats[key])
```
nmearl: Remove print statement.

specviz/plugins/statistics/statistics_widget.py
```diff
@@ -210,6 +259,38 @@ def clear_statistics(self):
         self._clear_stat_widgets()
         self.stats = None
 
+    def _reconnect_item_signals(self):
+        if self.hub.plot_item is self._current_plot_item:
+            return
+
+        if isinstance(self._current_plot_item, PlotDataItem):
```
nmearl: Just to clarify: is that a particular reason this was necessary? Are you expecting the plot item to be anything else?

specviz/plugins/statistics/statistics_widget.py
```diff
@@ -222,12 +303,19 @@ def update_statistics(self):
         spectral_region = self._get_workspace_region()
 
         self._current_spectrum = spec
+        self._reconnect_item_signals()
```
nmearl: I don't think you should be running this test every time the update statistics is called. Why not have the plugin listen for selection changes in the workspace, and update the current item reference there?

```python
# In your __init__
self.hub.workspace.current_selection_changed.connect(lambda plot_item: self.update_statistics())
```

The signal returns the currently selected item, which could be passed to the update statistics function.

specviz/plugins/statistics/statistics_widget.py
robelgeda: Nope, this is just a blanket check for `None` and anything that is not a `PlotDataItem` introduced in the future.

specviz/plugins/statistics/statistics_widget.py
robelgeda: We would still need a way to listen to unit changes. The main reason for this line is to make sure the stats widget updates when a unit changes. This is done by listening to `PlotDataItem.data_unit_changed` and `PlotDataItem.data_unit_changed`. I agree this is not the best way of doing it but there isn't a single signal for unit changes yet. Let me know if there is a better way to do this?

robelgeda:
@astrofrog @eteq @brechmos-stsci : @SaOgaz  and I spent quite a bit of time looking at the bug she mentioned. The reason the test is failing is because we are trying to deepcopy a `Spectrum1D` object initialized with `flux` and `wcs` kwargs. I tried to deepcopy a similar `Spectrum1D` in ipython and I get the same error. When I initialize with `flux` and `spectral_axis` only, it copies without an issue. If this is not unique to me, I can make a ticket.

nmearl:
I've put a fix for this in astropy/specutils/pull/390.

specviz/plugins/statistics/statistics_widget.py
SaOgaz: Sorry about this, I thought I had checkout a clean version of stuff, but clearly I hadn't 🤣

javerbukh:
Pulled this branch. Loaded Specviz with the COS_FUV.fits file, then added a region. I tried to resize the region but every pixel I moved came with this output in terminal `ERROR:root:Spectrum1D currently requires the uncertainty be defined.` It made the region moving extremely laggy, not sure if anyone else got this bug.

SaOgaz:
I've never seen the slow re-sizing problem.  The Error might be coming from one of the calculations.  @robelgeda, would you have time to look at this?

brechmos-stsci:
@javerbukh @SaOgaz The specutils SNR requires the uncertainty be defined in the Spectrum1D object. So, if it is not defined then you will get the error.  Specviz will have to decide what to do about it. There is a method in specutils to define an uncertainty based on an RMS type error of the spectrum, so that is one option.

@robelgeda

SaOgaz:
I'm pretty sure we decided that the no SNR calculation was acceptable, hence filling that box in with a NaN.  But, if the logging is slowing down the API, we should address that.

nluetzge:
Hmmm I used to to use this formular alot. It doesnt need uncertainies but I’m not sure how accurate it still is. Maybe @eteq or @gderosa2004 can weigh in? http://www.stecf.org/software/ASTROsoft/DER_SNR/

brechmos-stsci:
@nluetzge Maybe post an issue in specutils as an enhancement to add another SNR calculation method.

hcferguson:
That algorithm looks quite handy; agreed it's best to put in SpecUtils first. The link to the newsletter article (which justifies the algorithm) in the python script didn't work, but a little googling led to this https://www.spacetelescope.org/static/archives/stecfnewsletters/pdf/hst_stecf_0042.pdf.

nluetzge:
Ok cool. Will open an issue in specutils then.

Issue #479, opened 2018-10-24
specviz crash: model fitting


gderosa2004:
If I load a spectrum (e.g. example_stis.fits), select a region, click on add model and select Const1d specviz crashes with the following message

```
Traceback (most recent call last):
  File "/Users/gderosa/Desktop/specviz/specviz/plugins/model_editor/model_editor.py", line 41, in <lambda>
    action.triggered.connect(lambda x, m=v: self._add_fittable_model(m))
  File "/Users/gderosa/Desktop/specviz/specviz/plugins/model_editor/model_editor.py", line 88, in _add_fittable_model
    idx = self.model_tree_view.model().add_model(model())
AttributeError: 'NoneType' object has no attribute 'add_model'
Abort trap: 6
```

brechmos-stsci:
@robelgeda is working on the fitting part, so this one might have to wait and probably could be associated with https://github.com/spacetelescope/specviz/issues/414

nmearl:
This is actually due to the fact that a model-based spectrum object hasn't yet been created. That is, currently, you'd need to select the data in the list, click the "New Model" button at the top. This creates a new data item in the list that is model-based. You'd then add new sub models to this model-based spectrum.

nmearl:
Certainly the plugin should be disabled until the user has created a new model spectrum, or selects an already created one.

nmearl:
#500 now disallows users from preemptively adding submodels.

<img width="1392" alt="screen shot 2018-10-29 at 12 50 27 pm" src="https://user-images.githubusercontent.com/4141126/47666240-6bbeeb00-db79-11e8-84ab-c0640c2b9839.png">

hcferguson:
Addressed by #500.

PR #500, opened 2018-10-29
Disallow interaction until user has created a model

nmearl:
Addresses #479.

### DIFF
--- a/specviz/core/hub.py
+++ b/specviz/core/hub.py
@@ -1,4 +1,6 @@
-from qtpy.QtWidgets import QApplication
+import logging
+
+from .items import DataItem
 
 
 class Hub:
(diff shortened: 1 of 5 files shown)


javerbukh:
If I select the filename and click the new model button, Specviz crashes with the following traceback
```Traceback (most recent call last):
  File "/Users/javerbukh/Documents/specviz/specviz/plugins/model_editor/model_editor.py", line 89, in _create_new_model
    model_data_item.identifier)
  File "/Users/javerbukh/Documents/specviz/specviz/core/models.py", line 120, in item_from_id
    self._items[data_item.identifier] = PlotDataItem(data_item)
  File "/Users/javerbukh/Documents/specviz/specviz/core/items.py", line 75, in __init__
    self._data_unit = self._data_item.flux.unit.to_string()
  File "/Users/javerbukh/Documents/specviz/specviz/plugins/model_editor/items.py", line 18, in flux
    return result(self.spectral_axis.value) * self.data(self.DataRole).flux.unit
TypeError: 'NoneType' object is not callable
Abort trap: 6```

Issue #503, opened 2018-10-29
ASCII support


hcferguson:
This issue is to solidify what we should support for ASCII input files.

1. uncommented wavelength & flux (don't care about no stinkin' units):

```
4000 5.e-16
4001 5.1e-16
```

2. commented but with no units

```
# wave flux uncertainty 
4000 5.e-16 1.e-17
4001 5.1e-16 1.e-17
```

```
# wavelength flux uncertainty 
4000 5.e-16 1.e-17
4001 5.1e-16 1.e-17
```

3. IPAC format

```
\\name=value
\\ Comment
| wavelength | flux              | uncertainty       | mask |
| double     | double            | double            | int  |
| Angstrom   | erg cm-2 s-1 AA-1 | erg cm-2 s-1 AA-1 | None | 
| null       | null              | null              | null |

```

nmearl:
Should we assume some default units for the no units case and/or which units would be preferable?

hcferguson:
I would suggest defaulting to "pixels" and "counts" when no units are specified.

hcferguson:
For IPAC, the question is how fancy to get with allowing different names and associated units. For example would we allow someone to come in with "frequency" and the appropriate unit and parse that automatically? What about "freq"? 

For this first pass my inclination is to throw it all to the wizard and not try to put too much logic in the loader. We're not going to be getting too many spectra in IPAC format anyway. But the same question arises for ECSV and ASDF, so whatever we do we should have the same logic for all three.

nmearl:
The loaders parse by column, and assume the order of the columns with no regard for units -- i.e. first is dispersion (wavelength/frequency/velocity...), second is data, third uncertainty. The unit-ful arrays get passed to specutils, and specutils takes care of parsing the unit behavior.

If the first column is in frequency, `my_spec.spectral_axis` will be in frequency units, but `my_spec.wavelength` will be in Angstrom, for example.

PR #504, opened 2018-10-29
First pass on generic ascii/ipac loaders

nmearl:
Addresses: #503.


hcferguson:
I'm confused. I thought the loaders were going into SpecUtils and SpecViz would just call them?

nmearl:
I assumed because the issues were opened in the specviz repo, that these would be implemented here. It really doesn't matter, everything is specutils underneath. These can be copy-pasted to specutils in their entirety (there have been previous discussions about moving __all__ the specviz loaders to specutils eventually anyway).

PR #505, opened 2018-10-29
WIP: Make default line width 2 pixels

brechmos-stsci:
The default line width for the spectra is 1 pixel which is quite narrow.  Changed it to 2.

1 pixel width:
image

2 pixel width:
image

Talked to @nluetzge and @gderosa2004 and they were good with the increase in default line width.

### DIFF
--- a/specviz/core/items.py
+++ b/specviz/core/items.py
@@ -75,7 +75,7 @@ class PlotDataItem(pg.PlotDataItem):
         self._data_unit = self._data_item.flux.unit.to_string()
         self._spectral_axis_unit = self._data_item.spectral_axis.unit.to_string()
         self._color = color or next(flatui)
-        self._width = 1
+        self._width = 2
         self._visible = False
 
         # Include error bar item


nmearl:
Default should be 1 for performance reasons. For example, try loading and playing around with the `example_generic.fits` when the width is > 1.

brechmos-stsci:
Huh?  Display is affected by linewidth?  Yikes, that is not good.

brechmos-stsci:
Just tried it, that is horrible. I have a couple ideas, so @javerbukh hold off on this for a bit

nmearl:
Of course it is. You're increasing the rendering load every iteration of the event loop, just as in any live graphics rendering application.

There are optimization that can be done in pyqtgraph -- opengl enabled by default for gpu acceleration, decimation of the spectrum when the point density is higher than can be visually distinguished, view clipping, etc.

nmearl:
See [the pyqtgraph docs](http://www.pyqtgraph.org/documentation/graphicsItems/plotdataitem.html) for `autoDownsampling`. And the [`useOpenGL` flag](http://www.pyqtgraph.org/documentation/widgets/graphicsview.html#pyqtgraph.GraphicsView.__init__) for the graphics view (it's set in the options dict, though).

brechmos-stsci:
Closing this as this solution will not work with lines that have a large number of points.

PR #512, opened 2018-10-30
Baseline docs layout

eteq:
This PR creates the basic documentation structure for the "new specviz".

Note that the "getting-started" is intentionally un-populated.  This is meant as sort of an "example" of what these sections should have - follow-on PRs should both populate the getting started itself, and the more detailed sections on specific tasks.

I also used a lot of the install/launch instructions from "old specviz" - I think I cleaned it up to be all applicable still, but @nmearl may want to take a quick look to be sure.


eteq:
To say just a bit more here (prompted by https://github.com/spacetelescope/specviz/issues/514#issuecomment-434489041): the idea of this outline is that *many* more sections should be created in the same vein as the "getting stated" section.  That is, there should be separate pages for the major features.  I'm intentionally not adding them here because we don't *have* all of those features yet, so rather than creating half-empty docs, I created a template (the getting started section) and other PRs implementing the features should add the docs for that particular feature.

Issue #510, opened 2018-10-30
Plan for New File Loader Wizard


javerbukh:
This is a continuation of #416 and all discussion of Loader Wizard changes will happen here.

javerbukh:
1. Templates made to load a certain kind of file and how templates map to dictionary that is being output
2. New way of saving python template files. Update UI to reflect that python templates are used instead of YAML, and that a copy of those templates is always saved in .specutils. If a user is going to overwrite a filename, create dialog to warn them first. Decide what kind of button presses makes the most sense (OK, SAVE, SAVE AS, CANCEL, etc.).
3. Determine if JWST spectrum files contain WCS as opposed to dispersion arrays.
4. Implement WCS loading
5. ASDF loader (might be differed until after the October 2018 sprint)

drdavella:
Related issue in specutils: https://github.com/astropy/specutils/issues/370

SaOgaz:
Maybe this is better as an in person conversation, but what assumptions do we want to make about a potential ascii input as far as formatting?  Was planning on using astropy tables to read in the file, and I could see having a loader builder that could pass on any of the astropy Table pre-set formats through the wizard...

hcferguson:
@sogaz happy to chat about this.

SaOgaz:
@javerbukh @hcferguson , are we good to close this issue now?

hcferguson:
@hcferguson Yes, I think this can be closed.

SaOgaz:
I'm going to be opening new separate issues for the file loader wizard, so closing this one.

Issue #515, opened 2018-10-31
Philosophy for dealing with collections of spectra in a single input file


hcferguson:
We should decide on a general philosophy for whether or not to support multiple 1d spectra in a single input file with SpecViz. Given that we are packaging JWST data that way, we probably should support this, but it's going to complicate the loaders.

The JWST data for MOS and Grism will typically be packaged together in a single file. In the FITS versions of these files, they have a primary HDU, then one extension per extracted spectrum, where the spectrum is in a BINTABLE with Wavelength, Flux, Err, DQ and a bunch of other columns. Then there is a final ASDF extension. 

Of course MOSViz is the tool intended for these collections, but people will no doubt throw them
at SpecViz.

ibusko:
Some HST spectra are also delivered by the pipeline as multiple 1D entries in a single file. STIS echelle extracted spectra comes to mind. Unlike JWST, they pack all 1D spectra together in a single FITS extension, using array tables (FITS "3-D" table format) where each cell in a row/column position stores an array. I recall that some COS spectra also follow a similar pattern.

SaOgaz:
Right now in PR #517, I'm running a flatten on the FITS table data when building a wizard built fits type loader.  @gderosa2004 informally approved of doing this when I quickly showed it to her.

gderosa2004:
@hcferguson, please correct me if I am wrong, but I think the intrinsic difference for MOS, grism spectra vs STIS/COS spectra is that in the first case we are dealing with spectra of multiple sources and we would like to keep them separated, while for COS and STIS we are talking about different wavelength ranges of the same spectrum so my preference as a user would be to have them merged together. Does it make sense?

ibusko:
@gderosa2004 regarding the STIS and COS format, for the sake of illustration, here is how we handled the problem in the old Java specview application:

The user was able to handle each one of the segments as a separate spectrum in itself, as a first take. Because merging would require interpolation and possibly normalization adjustments that would be outside the scope of a data loader. This is particularly true for the STIS echelle format, where each order partially overlaps the previous and next orders and the overlapping regions need to be combined together.  

The loader, by default, just created a kind of compound spectrum made of separate segments and delivered that to the caller. In one option, the user would be able to just overplot all these separate, individual segments on a single plot, thus generating a sort of "visual combination". Internally, the data would still be held in separate spectrum instances. Operations such as integrations or fit over wavelength ranges would have to deal with all the intrincacies caused by the overlaps. 

[...]

PR #517, opened 2018-10-31
Loader wizard refactor

SaOgaz:
Lots left to do on this, running out of time today, will document more tomorrow. This is mostly so @javerbukh can pull it tomorrow morning to start on the GUI end.

This is for #510


SaOgaz:
Current issues/questions:
- Still haven't done the ascii or ecsv style loaders.
- Still need to hook in WCS.

- Had to use a flatten on the data columns for the COS FUV data I was testing the fits style loaders with. Is this generally safe?
- Reload for new loaders isn't currently working, waiting on #370.
- Possible updates depending on #515 
- Right now I couldn't think of a way to avoid having a separate template for uncertainty.  You cannot pass the uncertainty argument in `Spectrum1D` `None`, which was the initial way I was going to work around it. It's not the end of the world to have multiple templates, but hoping they don't start multiplying out of control at some point.

SaOgaz:
Okay, here's the status.....

File types that seem to work okay with the wizard are:
COS (tried FUV and NUV)
simple ascii files
deimos
SDSS

Types that are not working:
image data (STIS, MDM, apogee(?))
STIS


This PR still needs:
- Get rid of the top set of buttons for each section (the which table is blah stored in), since that is handled indirectly through the table.read call
- Add a warning at the top of the wizard that it can only handle tabular data right now
- Have the wizard test out the loader it will build, make sure it will run without crashing on the data used to start the wizard
- We need to do something about the units fields, it will let you save it if it's blank, or none valid, which results in a not working loader.
- Write test/tests
- Write documentation
- Test ecsv
- Fix the following bugs:
   - "custom" units getting autochosen for dispersion and data sections
   - registration issue, code to remove and re-add the loader once the wizard is done saving it out isn't working properly.
   - uncertainty and masks "component" drop downs not refreshing with "refresh data" button click, unless you enable/disable/reenable those sections.

specviz/plugins/loader_wizard/loader_wizard.py
```diff
@@ -0,0 +1,631 @@
...
+
+        filename = "{}.py".format(filename) if not filename.endswith(".py") else filename
+
+        string = self.as_new_loader()
+        with open(filename, 'w') as f:
+            f.write(string)
+
+        # If a loader by this name exists, delete it
+        if self.new_loader_dict['name'] in registry.get_formats()['Format']:
+            registry.unregister_reader(self.new_loader_dict['name'], Spectrum1D)
```
drdavella: We also need to add the following here:

```python
registry.unregister_identifier(self.new_loader_dict['name'], Spectrum1D)
```

specviz/plugins/loader_wizard/loader_wizard.py
```diff
@@ -0,0 +1,631 @@
...
+
+        # Add new loader to registry
+        spec = importlib.util.spec_from_file_location(os.path.basename(filename)[:-3], filename)
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)
```
drdavella: Instead of these three lines, you can just do the following:

```python
specutils.io.registers._loader_user_io()
```

You'll need to add `import specutils` at the top of the module.

SaOgaz:
@nluetzge, this PR, although not 100% done, is ready for user feedback/testing!  I'd really like to know if the button layout / setup makes sense or if we should tweak it.

SaOgaz:
Note: we're currently flattening the input columns (data).  This may not be appropriate for certain spectra (COS?) that contain disjoint spectra in a two plus dimensional array in one table column.

nluetzge:
Pretty swamped today and tomorrow but I will see if I can have a look at it.

specviz/plugins/loader_wizard/loader_wizard.py
```diff
@@ -0,0 +1,646 @@
...
+            self.new_loader_dict['uncertainty_type'] = self.ui.combo_uncertainty_type.currentData()
+
+
+class ASCIIImportWizard(BaseImportWizard):
+    dataset_label = 'Table'
+
+    def as_new_loader_dict(self, name=None):
+        """
+        Convert the current configuration to a dictionary
+        that can then be serialized to YAML
```
nmearl:
```suggestion
        that can then be serialized to a Python file.
```

SaOgaz:
@nmearl, definitely agree with you on the jinja... we just didn't have time to implement it for this release cycle. I didn't know I could hide elements, I'll have to do that for the bit masking. Will address rest of your feedback as well, thanks for the review!

hcferguson:
Regarding exposing the Table.read() interface....we had talked about making that a dropdown of allowed formats, but that seemed too complicated to implement for this release.

SaOgaz:
If there's still time before the delivery, @drdavella offered to help me work out the drop down on monday, so I don't spin my wheels on it while trying to figure it out.  I've just about finished the changes requested by @nmearl, about to push.

SaOgaz:
Need to update the test url here, I can't figure out how to pull the api download url from box...

SaOgaz:
Bug list for issue once this get's merged:

- The dropdowns sometimes  get selected/re-selected oddly. Only happens occasionally, not sure under what circumstances. We should figure out what we want the default behaviour to be.

- Possible bug with unit fields re-appearing when table reload happens. Again, hard to re-create, need to figure out trigger.

- Need to re-add the close and okay buttons.  The okay button will load the data with the current wizard setup, but not save the loader out to a file.

- Are we okay with the default enter button attachment, might want to explore better options. Although at first glance the internet says it may not be possible to have no button attachment for the enter key.

nmearl:
@SaOgaz there's no way to accept the loader immediately? You have to save the file out, close the loader wizard, and re-oped using the regular file load dialog?

nmearl:
Also, the unit fields are still mis-aligned:

screen shot 2018-11-21 at 4 55 00 pm

How they should be:

screen shot 2018-11-21 at 4 57 16 pm

SaOgaz:
@nmearl, yeah I was having problems with designer, the drop downs above the drop down are stuck in the middle divider and I couldn't get them to cross the vertical section/grid line... any tips?

The okay/close buttons need to be re-added and the okay set to load the data.  @eteq, @gderosa2004, and I tagged this as a "it shouldn't hold up release" change.  But maybe others feel differently?  I don't know if I'll have time to work on it today, I have some other stuff that really needs my attention.

PR #523, opened 2018-11-01
Disable auto-si-prefixing on data axis

### LINKED ISSUE #522: flux units should not be auto-converted to SI prefixes by pyqtgraph
Most of the context for this issue is in #521, https://github.com/spacetelescope/specviz/pull/480#issuecomment-435141361 and https://github.com/spacetelescope/specviz/pull/480#issuecomment-435163125 .  Essentially the problem is this: if I load in a spectrum with flux of `array([4.0115904e-14, 2.7294973e-14, 2.5229817e-14, ..., 2.7819518e-14,
       1.6174598e-13, 1.6174598e-13], dtype=float32)`  and units of `'erg /s /cm**2 /angstrom'`, spcviz pops up with flux units of femto-erg /s /cm**2 /angstrom (a unit astronomers do not use, even though it is technically a valid unit).  That's not desirable behavior, because it hides from the user what the "true" values of the spectrum are.  Given one of specviz's use cases as a "quick look" tool, it's very important the user sees exactly what is in the file they load.

Of course, they should have the *option* to change it to another unit, which is what #480 enables.  That way they can *ask* to use femto-ergs. 

If I understand right from @nmearl this is a pretty trivial change of just turning off the automatic pyqtgraph auto-SI setting? (I scoped it 1 under that assumption)

nmearl:
Fixes #522.

### DIFF
--- a/specviz/widgets/plotting.py
+++ b/specviz/widgets/plotting.py
@@ -169,6 +169,7 @@ class PlotWidget(pg.PlotWidget):
         self.addItem(self._region_text_item, ignoreBounds=True)
         self._region_text_item.setParentItem(self.getViewBox())
         self.getAxis('bottom').enableAutoSIPrefix(False)
+        self.getAxis('left').enableAutoSIPrefix(False)
 
         # Store the unit information for this plot. This is defined by the
         # first data set that gets plotted. All other data sets will attempt


javerbukh:
Made outdated by the merge of #480, closing now.

PR #527, opened 2018-11-02
Loader ux improvements

nmearl:
Related to astropy/specutils#376. Moves all loaders to specutils; supports extensions and priorities in loaders; enables the functionality of auto-loading based on priorities and best match for identifier functions.


specviz/core/items.py
```diff
@@ -143,7 +142,8 @@ def is_data_unit_compatible(self, unit):
         return (self.data_item.flux.unit == "" or
                 unit is not None and
                 self.data_item.flux.unit.is_equivalent(
-                    unit, equivalencies=spectral_density(self.spectral_axis*u.Unit(self.spectral_axis_unit))))
+                    unit, equivalencies=spectral_density(
+                        self.data_item.spectral_axis)))
```
drdavella: Just curious but is this required for this PR, or is it unrelated cleanup?

specviz/widgets/workspace.py
```diff
@@ -354,8 +366,30 @@ def load_data(self, file_path, file_loader, display=False):
...
+                # loaders that fit the same identifier, choose the highest
+                # priority one.
+                fmts = identify_format('read', Spectrum1D, file_path, None, [], {})
+
+                logging.warning(e)
+                logging.warning("Loaders for '%s' matched for this data set. "
+                                "Iterating based on priority."
+                                "", ', '.join(fmts))
+
+                for fmt in fmts:
```
drdavella: Just a nit, but it would be useful to have a comment here that says the formats are already in priority order. It's not entirely obvious from the code (at least without looking in `specutils`).

PR #534, opened 2018-11-05
[WIP] Store reference to plugin in tool bar actions

nmearl:
This stores a simple reference to a plugin action's parent plugin object class so that it can be accessed via the gui testing functions.

### DIFF
--- a/specviz/core/plugin.py
+++ b/specviz/core/plugin.py
@@ -56,13 +56,6 @@ class DecoratorRegistry:
 
 
 class Plugin(DecoratorRegistry):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        self._plugin_bar_decorator = PluginBarDecorator()
-        self._tool_bar_decorator = ToolBarDecorator()
-        self._plot_bar_decorator = PlotBarDecorator()
-
     @property
     def registry(self):
         return self._registry
@@ -89,6 +82,8 @@ class Plugin(DecoratorRegistry):
                 for meth_name, meth in members:
                     if hasattr(meth, 'wrapped') and (filt is None or meth.plugin_type == filt):
                         meth(workspace)
+
+                return plugin
 
             self._registry.append(cls_wrapper)
 
@@ -162,10 +157,6 @@ class Plugin(DecoratorRegistry):
                 action = QAction(parent)
                 action.setText(name)
 
-                # Store a reference to the plugin in the ui-added action
-                # TODO: come up with better solution...
-                action.plugin = plugin
-
                 if icon is not None:
                     action.setIcon(icon)
 
@@ -198,10 +189,6 @@ class Plugin(DecoratorRegistry):
                 parent = workspace.current_plot_window.tool_bar
                 action = QAction(parent)
 
-                # Store a reference to the plugin in the ui-added action
-                # TODO: come up with better solution...
-                action.plugin = plugin
-
                 action.setText(name)
 
                 if icon is not None:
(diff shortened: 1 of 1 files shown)


javerbukh:
In the case of unit conversion, the class uses exec_() to run. Does this mean I will need to use signals and slots in my testing?

@nmearl @drdavella

drdavella:
@javerbukh as we discussed over slack, I think that plugin widgets should be using `show` instead of `exec_` in order to allow tests to maintain control flow. There are numerous examples of this in the tools in `cubeviz`. For example, consider the `display` method implemented by the moment maps GUI:

https://github.com/spacetelescope/cubeviz/blob/f0c14bdb114e388cdd24878ba22f9fd43781479c/cubeviz/tools/moment_maps.py#L27-L88

nmearl:
@drdavella I'm not sure I follow. The `Plugin` class is not a part of a plugin's inheritance hierarchy (it's just the registry) -- the plugin is whatever class is wrapped by the plugin decorator. What you're describing would be (if I'm understanding what you're saying), just storing a reference to itself.

The problem is that toolbar actions are generally the things exposed to the ui from a plugin. These actions don't currently contain information about which plugin they belong to, so getting the `Smoothing` plugin toolbar action doesn't provided a reference to the actual `SmoothingDialog` class. That's what this PR does.

drdavella:
@nmearl the problem currently is that since we wrap the class with a decorator, trying to directly instantiate `UnitChangeDialog`, for example, is not possible. The new object gets swallowed by the decorator and is never returned to the calling code.

I'm proposing that we actually return the resulting object so that tests can instantiate plugins like `UnitChangeDialog` directly instead of having to click for them in the toolbar, which is quite awkward and doesn't really test functionality we care about.

nmearl:
Ah, that's a good point. However, in python 3, the wrapped function or class is always accessible by the `__wrapped__` attribute:

```
unit_change_dialog = UnitChangeDialog.__wrapped__()
```

This avoids triggering the `Plugin` decorator. This would certainly allow tests to occur without having to interact with the toolbar actions.

nmearl:
However, it should be noted that instantiating the class by itself does not then inject a reference to the `Hub` class, and, by extension, to the internal data model, which is required for many plugins. I think ideally, we need to setup a fixed testing environment through which pytest can provided `Hub` stubs and pre-populated data items for use in testing. But that might be beyond the scope of this sprint for now.

drdavella:
@nmearl that makes sense. In `cubeviz` it was sufficient to pass the application instance (or something similar) to the widget constructor in order to provide access to all of the relevant state. I was hoping it might be a similar situation here, but it seems a bit more complicated.

I wonder if there's a middle ground: can we make these plugins a bit more accessible from the workspace instead of having to find them in the toolbar?

drdavella:
@nearl, just to clarify, if I do something like the following (given the change I proposed above), is it still not enough?

```python
from specutils.app import Application
from specviz.plugins.unit_change.unit_change_dialog import UnitChangeDialog

app = Application([], dev=True)
ucd = UnitChangeDialog(app.current_workspace)
```

I was hoping this was all that would be required for testing something like the unit change UI.

nmearl:
Alright, I believe the below should work to setup some developer test data and give access to the plugin within the specviz environment.

```python
from specviz.app import Application
from specviz.plugins.unit_change.unit_change_dialog import UnitChangeDialog

app = Application([], skip_splash=True, dev=True)

ucd = UnitChangeDialog(app.current_workspace)

ucd.show()

...

ucd.close()
```

nmearl:
@drdavella That should work, I'd suggest adding `skip_splash` to the `Application` keyword, but, yeah, that seems right.

specviz/core/plugin.py
```diff
@@ -56,13 +56,6 @@ def get_action(parent, level=None):
 
 
 class Plugin(DecoratorRegistry):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        self._plugin_bar_decorator = PluginBarDecorator()
-        self._tool_bar_decorator = ToolBarDecorator()
-        self._plot_bar_decorator = PlotBarDecorator()
```
drdavella: Was all of this just extraneous?

nmearl: Yep. I'm quite sure that I actually made the change to remove this stuff in an earlier PR that I closed for some reason, and forgot to make the changes to a later PR.

PR #537, opened 2018-11-06
Load multiple spectra from a single file

drdavella:
This builds on the work done in https://github.com/astropy/specutils/pull/381 in order to allow files that contain multiple heterogenous spectra to be loaded.

image

This PR should not be merged until the above PR in `specutils` is accepted and merged. Also I need to add some tests here. I plan to create mock data files on the fly and simply test the backend loader functionality (i.e. ignoring the UI).

Also this will need to be rebased on #527 once it is reviewed and merged.

### DIFF
--- a/specviz/widgets/workspace.py
+++ b/specviz/widgets/workspace.py
@@ -437,6 +437,17 @@ class Workspace(QMainWindow):
 
                 message_box.exec()
 
+    def _add_and_plot_data(self, spectrum, name):
+        data_item = self.model.add_data(spectrum, name=name)
+
+        # If there are any current plots, attempt to add the data to the plot
+        plot_data_item = self.proxy_model.item_from_id(data_item.identifier)
+        plot_data_item.visible = True
+        self.current_plot_window.plot_widget.on_item_changed(data_item)
+        self._on_item_changed(item=plot_data_item.data_item)
+
+        return data_item
+
     def load_data(self, file_path, file_loader=None, display=False):
         """
         Load spectral data given file path and loader.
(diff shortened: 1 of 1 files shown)


drdavella:
This requires a minor but significant patch to `specutils` in order to work: https://github.com/astropy/specutils/pull/396

specviz/widgets/workspace.py
```diff
@@ -354,6 +354,17 @@ def compose_filter_string(reader):
 
         self.load_data(file_path, file_loader=loader_name_map[fmt])
 
+    def _add_and_plot_data(self, spectrum, name):
```
nmearl: There already is a method `force_plot` that does what this method does. Is there are reason to duplicate that functionality?

specviz/widgets/workspace.py
```diff
@@ -320,8 +320,8 @@ def _on_load_data(self):
         """
         When the user loads a data file, this method is triggered. It provides
         a file open dialog and from the dialog attempts to create a new
-        :class:`~specutils.Spectrum1D` object and thereafter adds it to the
-        data model.
+        :class:`~specutils.SpectrumList` object and thereafter adds the
+        contents to the data model.
```
nmearl: It's not adding `SpectrumList` to the data model -- it's slicing the `SpectrumList` to get `Spectrum1D`s, and adding them to the data model. May want to just clarify that?

specviz/widgets/workspace.py
drdavella: This was a leftover from the pre-rebase work that I did. This function is slightly different but can be refactored in terms of `force_plot`.

specviz/widgets/workspace.py
drdavella: Yep I think this may have been the result of a blind search-and-replace.

drdavella:
Just pushed to resolve conflicts. Once CI passes this should be ready to merge.

PR #538, opened 2018-11-06
Fix bug with wave range dialog when working with multiple plot windows.

ibusko:


specviz/widgets/linelists_window.py
```diff
@@ -113,6 +113,9 @@ def __init__(self, plot_window, parent=None):
 
         self.plot_window = plot_window
 
+        global wave_range
```
nmearl: Is there a reason that the `wave_range` variable is module-level? Can it be incorporated as a class-level variable in `LineListsWindow`? I only ask because globals are generally a way to share information in more functional-style programing instead of object-oriented, and due to several issues, are generally recommended against using unless absolutely necessary.

ibusko: Indeed, there is no reason. wave_range can be now just an instance variable. This is probably a leftover from older versions. I added a comit that fixes the code.

PR #540, opened 2018-11-07
Old: Unit conversion tests and code changes

javerbukh:
Conftest.py now loads in dev mode with fake data and skips the splash screen. Unit change dialog now uses show() instead of exec_(). Test ucgui consists of 8 tests that check different facets of uc dialog functionality.

### DIFF
--- a/specviz/plugins/unit_change/tests/test_unit_conversion_gui.py
+++ b/specviz/plugins/unit_change/tests/test_unit_conversion_gui.py
@@ -5,6 +5,22 @@ from qtpy.QtWidgets import QMainWindow, QMdiSubWindow, QListWidget, QAction, QDi
 
 from ..unit_change_dialog import UnitChangeDialog
 
+from specviz.plugins.unit_change.unit_change_dialog import UnitChangeDialog
+
+def test_ucd(specviz_gui, qtbot):
+    ucd = UnitChangeDialog(specviz_gui.current_workspace)
+    uc = ucd
+    uc.show()
+    qtbot.addWidget(uc)
+
+    assert uc.ui.comboBox_spectral.currentText() == "Angstrom"
+
+    uc.ui.comboBox_spectral.setCurrentIndex(uc.ui.comboBox_spectral.count()-1)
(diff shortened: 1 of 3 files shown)


specviz/plugins/unit_change/tests/test_unit_conversion_gui.py
```diff
@@ -1,59 +1,137 @@
...
-#     assert uc.current_units == old_units
+from qtpy.QtWidgets import QDialogButtonBox
+
+from specviz.plugins.unit_change.unit_change_dialog import UnitChangeDialog
+
+
+def get_ucd(specviz_gui):
+    ucd = UnitChangeDialog(specviz_gui.current_workspace)
+    ucd.show()
+    return ucd
```
drdavella: @javerbukh I've made yet another minor change to this in #544, so ultimately I think this function will end up looking like the following:

```python
def get_ucd(specviz_gui):
    ucd = specviz_gui.current_workspace._plugins['Unit Change Dialog']
    ucd.show()
    return ucd
```

This will have the advantage of not adding another icon to the workspace toolbar. I don't think this needs to be changed before merging, though.

specviz/plugins/unit_change/tests/test_unit_conversion_gui.py
```diff
@@ -1,59 +1,137 @@
...
+    """
+    Accept works for units in combobox
+    """
+    ucd = get_ucd(specviz_gui)
+    qtbot.addWidget(ucd)
+    
+    if ucd.ui.comboBox_spectral.count() > 4:
+        ucd.ui.comboBox_spectral.setCurrentIndex(4)
+        assert ucd.ui.comboBox_spectral.currentText() == ucd.spectral_axis_unit_equivalencies_titles[4]
+    else:
```
drdavella: Under what conditions would either of these be executed? I'm a little uncomfortable with tests that could have different results on different runs.

nmearl: Agree with @drdavella: this should only be testing a single, expected value.

javerbukh:
Overtaken by #547

Issue #542, opened 2018-11-07
Add tests and documentation for arithmetic GUI


drdavella:
The GUI looks good but we need a way to test it.

javerbukh:
Solved by #544.

PR #544, opened 2018-11-07
Add basic test for arithmetic gui

drdavella:
This partially addresses #542. It also fixes a minor bug, although more discussion might be required with @mfixstsci and @nmearl.

I hope that this serves as a good template for additional testing.

@nmearl as you can see here I actually added a pointer to the plugins in the workspace in order to facilitate testing. The change we made in #534 was useful, but creating a new instance of a plugin has the side effect of adding another icon to the toolbar, which I'd like to avoid. I think what I've done here is reasonable, though.

### DIFF
--- a/specviz/conftest.py
+++ b/specviz/conftest.py
@@ -64,5 +64,5 @@ def specviz_gui():
     for testing of specviz gui
     :return:
     """
-    spec_app = Application([])
+    spec_app = Application([], skip_splash=True, dev=True)
     return spec_app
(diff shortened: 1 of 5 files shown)


drdavella:
Also I removed extraneous whitespace from the arithmetic gui module. We should probably not allow failures on our `pycodestyle` test.

specviz/conftest.py
```diff
@@ -64,5 +64,5 @@ def specviz_gui():
     for testing of specviz gui
     :return:
```
nmearl: @drdavella since we're talking about testing coding style, docstrings should be numpy-styled.

PR #543, opened 2018-11-07
Model Fitting

robelgeda:
Closes #414 
This PR enables model fitting via SpecUtil fit_lines function. **For testing**, Please make sure to update your SpecUtils.
screen shot 2018-11-08 at 4 20 53 am

PR #552, opened 2018-11-09
Add model saving and loading

### LINKED ISSUE #516: Model saving

drdavella:
Currently this just uses `pickle` for simplicity, but it should be easy to adapt for other output formats.

This resolves #516.

Also, this is based on #543 and will need to be rebased once that is merged.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,7 +25,6 @@ Using SpecViz
 
    getting-started
    unit_conversion
-   model_fitting.rst
    arithmetic-layer
    model_fitting
    stats_sidebar
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,6 +25,7 @@ Using SpecViz
 
    getting-started
    unit_conversion
+   model_fitting.rst
    arithmetic-layer
    model_fitting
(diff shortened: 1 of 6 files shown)


docs/index.rst
```diff
@@ -25,6 +25,7 @@ Using SpecViz
 
    getting-started
    unit_conversion
+   model_fitting.rst
    arithmetic-layer
    model_fitting
```
robelgeda: There are two model fitting entries, was this on purpose?

specviz/plugins/model_editor/model_editor.py
```diff
@@ -150,27 +160,72 @@ def _on_remove_model(self):
...
+
+        self._load_model_from_file(filename)
+
+    def _add_model(self, model):
+        idx = self.model_tree_view.model().add_model(model)
+        self.model_tree_view.setExpanded(idx, True)
+
+        for i in range(0, 4):
+            self.model_tree_view.resizeColumnToContents(i)
+
```
robelgeda: we should probably add a `self._redraw_model` call here

PR #557, opened 2018-11-13
Model Editor Checkbox Fix

robelgeda:
Model editor's `Fixed` column crashes when the checkbox is clicked. This PR fixes this bug. I also added "Plot Units" to every unit display in the model editor param box. Lastly, I made sure that the model editor redraws a plot only if the selected `plot_data_item` is of type `ModelDataItem`.

### DIFF
--- a/specviz/plugins/model_editor/items.py
+++ b/specviz/plugins/model_editor/items.py
@@ -23,6 +23,11 @@ class ModelDataItem(DataItem):
         return self.data(self.DataRole).flux
 
     @property
+    def spectrum(self):
+        flux = self.flux  # Update the flux
+        return super().spectrum
+
+    @property
     def model_editor_model(self):
         return self._model_editor_model
 
--- a/specviz/plugins/model_editor/model_editor.py
+++ b/specviz/plugins/model_editor/model_editor.py
@@ -192,7 +192,8 @@ class ModelEditor(QWidget):
                 spectrum = spectrum.new_flux_unit(plot_data_item.data_unit)
                 plot_data_item.data_item.set_data(spectrum)
 
-        plot_data_item.set_data()
+            # Only draw if ModelDataItem
+            plot_data_item.set_data()
 
     def _on_model_item_changed(self, item):
         if item.parent():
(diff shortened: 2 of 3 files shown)


specviz/plugins/model_editor/items.py
```diff
@@ -22,6 +22,11 @@ def flux(self):
 
         return self.data(self.DataRole).flux
 
+    @property
+    def spectrum(self):
+        flux = self.flux  # Update the flux
+        return super().spectrum
+
```
drdavella: Is this change related to the rest of the PR?

specviz/plugins/model_editor/model_editor.py
```diff
@@ -192,17 +192,19 @@ def _redraw_model(self):
...
             # If the item has a parent, then we know that the parameter
             # value has changed. Note that the internal stored data has not
             # been truncated at all, only the displayed text value. All fitting
             # uses the full, un-truncated data value.
-            item.setData(float(item.text()), Qt.UserRole + 1)
-            # item.setText("{:.5g}".format(float(item.text()))) # dont change user input
-            item.setText(item.text())
+            if item.column() == 1:
+                item.setData(float(item.text()), Qt.UserRole + 1)
+                # item.setText("{:.5g}".format(float(item.text()))) # dont change user input
```
drdavella: Should remove this comment if it's no longer useful.

specviz/plugins/model_editor/models.py
```diff
@@ -77,7 +77,8 @@ def add_model(self, model):
             param_value.setData(parameter.value, Qt.UserRole + 1)
 
             # Store the unit information
-            param_unit = QStandardItem("{}".format(parameter.unit))
+            # param_unit = QStandardItem("{}".format(parameter.unit))
```
drdavella: Should remove this comment if it's no longer useful.

robelgeda: We need the commented out lines when units are enabled. They are there because those lines are not final.

specviz/plugins/model_editor/model_editor.py
robelgeda: This line is not final. There needs to be a discussion on how to display the fitted params.

specviz/plugins/model_editor/items.py
robelgeda: This is related to the 

> I made sure that the model editor redraws a plot only if the selected plot_data_item is of type ModelDataItem.

part. It ensures that when the redraw function replaces the Spectrum1D object in this data item, the data item does not return a Spectrum1D with the wrong flux values. The `flux = self.flux` line updates the flux values of the Spectrum1D by overriding its `_data` attribute (see the `flux` property for how it does this).

Issue #573, opened 2018-11-20
Have a concept of "the continuum" in specviz and use it in the statistics plugin


eteq:
This is a discussion item for a future version (timeline TBD).

I think it might make sense with many of the workflows discussed for there to be a formal concept of "the continuum".  This could be as simple as a special singleton model, identified by a special name ("Continuum"?) that if present, is interpreted as the continuum.  This model could be auto-generated using the `specutils` continuum-fitting functions, but then it would show up just as another model in the model editor.

The idea would then be for the panel being re-worked in #572 to be "smart" about it's contents.  That is, instead of the #572 result that it works on just whatever spectrum is currently active, the "cont sub" and "cont norm" parts of that plugin could *automatically* do the subtraction/normalization.  That I think makes more sense with the multi-tab layout, anyway.

nmearl:
I'm not a big fan of having things implicitly generate itself. I'd rather have something like a function in the 'Operations' menu -- 'Auto-generate Continuum Model' -- that will do a "best-guess" at the continuum, and add it to the data list as a regular model data item that the user can tweak if they want.

Likewise, I think that the statistics should have a drop down to define which data item the user wants to the statistics to use as the continuum. I can imagine the work flow where the user has done some arithmetic operations and the spectrum they are interested in is no longer the spectrum that the auto-generated continuum model fits, so they create a new auto-generate continuum from this new spectrum. However, the original spectrum and its original auto-generated continuum model are still available, and the user can choose which data item represents the continuum for their statistics.

hcferguson:
Closing this as addressed in #574. Starting a wish-list of continuum fitting enhancements in #603.

Issue #533, opened 2018-11-05
Continuum fitting


hcferguson:
Continuum fitting tends to be a combination of fitting and filtering. As such it is conceptually slightly different from model fitting, although can use most of the same machinery. 

Astronomers will be comfortable having a toolbox (or plugin) for "fitting continuum". SpecUtils now has a continuum.py module, so the task is basically to expose that in SpecViz.

hcferguson:
Closing this since #574 added basic functionality. Will address specific fixes/enhancements in separate issues.

PR #574, opened 2018-11-21
Auto generate continuum

nmearl:
Added a new operation that will auto-generate a continuum data item and associated model for a given spectrum.

Related to #573 and #533.


nmearl:
Try now?

drdavella:
Much better now, thanks!

PR #577, opened 2018-11-26
Model Editor UI Upgrade

robelgeda:
The following updates are included in this PR:

- Rearrange the model editor UI buttons.

- Add tooltips to tool buttons

- Update `Fit Model` button (blue and larger).

- Add a load model button. 

- Model data item remembers which data item was fitted such that when the user goes through multiple models, the data selector combo box at the bottom is updated to show which data was fitted by that model data item.

- If there are no models in the model editor, set model flux values to 0.

- When the user has nothing plotted, they should be able to select any data item to fit... even if the new data item has incompatible units as the model's current plot data item. This PR introduces a snippet that changes the units of the model plot items in each plot window when the user selects an incompatible data item from the data selection combo box.

screen shot 2018-11-26 at 10 56 26 am

CC: @hcferguson @eteq @nmearl


nmearl:
Should the button for advanced fitting settings be down near the model/fitting stuff?

robelgeda:
@nmearl We could put it at the bottom. I wanted to put global model editor buttons on top and buttons affecting the current model at the bottom. We could do the following if its better:
screen shot 2018-11-26 at 11 35 45 am

hcferguson:
@nmearl, @robelgeda -- I don't have strong feelings on layout here, and hope we can get a UI expert in to help in the future. Happy to go with whatever requires the least work to get this merged.

robelgeda:
@nmearl ?

robelgeda:
Reorganized things a bit more:
screen shot 2018-11-27 at 10 46 04 am

PR #578, opened 2018-11-26
Statistics Type Selection

### LINKED ISSUE #572: Change to "statistics" plugin to better represent user workflows
As discussed in the specviz tag-up today, we had several changes to make to the "Statistics" plugin to make it more appropriate for typical astronomer workflows.

The primary issue is that the "statistics" don't all make sense because, e.g. Equivalent Width makes no sense for a non-continuum-normalized spectrum. 

While a longer-term improvement might be to keep track of continuum operations and be smart about what you show, for now a faster solution is to split the statistics out into several sub-panels. (This could be distinct plugins, although @nmearl is opposed to that in favor of sub-panels inside the "statistics" plugin).  These panels would be "Statistics" (or maybe "Generic statistics"?), "Continuum Subtracted", and "Continuum Normalized".  The continuum ones could just have some text above that says something like "this is only meaningful if you've done the right thing with the continuum". 

More concretely, the following would be "generic":
* Mean
* Median
* Std Dev
* RMS (although see below)
* Max
* Min
* Total

The following would be continuum-sub:
* Centroid
* FWHM
* Line Flux (does not yet exist, but needs to be added)

And this would be the only one that continuum-normalized:
* Equivalent Width 

Now some questions/suggests that at least I have:
* Why is there both RMS and std dev?
* Why is this called "statistics"?  A flux/centroid/eqw is usually thought of as a "measurement" rather than a "statistic".  Perhaps the more generic term might be just "analysis"? (In analogy to the specutils sub-package)

This suggests to me the following actions are required here:
- [x] Add the new panels with intermediate text that indicates they are only meaningful if the spectrum is continuum-whatevered
- [x] Move the various statistics into the right panels
- [ ] Add line flux
- [x] Add text at the top of the panel that says "these statistics are applied to the currently-selected region"

cc @SaOgaz @hcferguson @javerbukh @ibusko @robelgeda as participants in this discussion

robelgeda:
This PR introduces the following 

- Closes #572 : Add combo box to select stats type as discussed in #572 

- Adds `plot_window_activated` signal that fires when a new plot sub-window is opened. This fixes the lag that occurs when moving an ROI. i.e: the signal I used previously to listen to plot window activations was also firing for other events.

- Update the ROI info box to indicate that its stats over a single ROI. Improve range labels of the ROI.

### DIFF
--- a/specviz/plugins/statistics/statistics_widget.py
+++ b/specviz/plugins/statistics/statistics_widget.py
@@ -65,12 +65,6 @@ def compute_stats(spectrum):
         cent = "Error"
 
     try:
-        rms = np.sqrt(spectrum.flux.dot(spectrum.flux) / len(spectrum.flux))
-    except Exception as e:
-        logging.debug(e)
-        rms = "Error"
-
-    try:
         snr_val = snr(spectrum)
     except Exception as e:
         logging.debug(e)
@@ -98,7 +92,6 @@ def compute_stats(spectrum):
             'median': np.median(spectrum.flux),
             'stddev': spectrum.flux.std(),
             'centroid': cent,
-            'rms': rms,
             'snr': snr_val,
             'fwhm': fwhm_val,
             'ew': ew,
@@ -150,7 +143,6 @@ class StatisticsWidget(QWidget):
             'median': self.median_text_edit,
             'stddev': self.std_dev_text_edit,
             'centroid': self.centroid_text_edit,
-            'rms': self.rms_text_edit,
             'snr': self.snr_text_edit,
             'fwhm': self.fwhm_text_edit,
             'ew': self.eqwidth_text_edit,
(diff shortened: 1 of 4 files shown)


specviz/plugins/statistics/statistics_widget.py
```diff
@@ -168,6 +172,9 @@ def _init_ui(self):
                         (doc.documentMargin() + widget.frameWidth()) * 2 +
                         margins.top() + margins.bottom())
             widget.setFixedHeight(n_height)
+        self.comboBox.addItems(["Generic", "Continuum Subtracted", "Continuum Normalized"])
```
nmearl: Just a suggestion: maybe "Basic" instead of "Generic" -- not sure what "generic" statistics are. 😝

specviz/plugins/statistics/statistics_widget.py
```diff
@@ -348,6 +378,7 @@ def update_statistics(self):
         self.stats = compute_stats(spec)
         self._update_stat_widgets(self.stats)
         self.set_status(self._get_target_name())
+        print("self.sender()", self.sender())
```
nmearl:
```suggestion
```

specviz/widgets/workspace.py
```diff
@@ -302,6 +303,9 @@ def _on_sub_window_activated(self, window):
         # Re-evaluate plot unit compatibilities
         window.plot_widget.check_plot_compatibility()
 
+        # Fire a signal letting everyone know a new plot window has been added
```
nmearl:
```suggestion
        # Fire a signal letting everyone know a plot window has been activated
```

specviz/plugins/statistics/statistics_widget.py
robelgeda: Changed in `b64d35d`

PR #579, opened 2018-11-26
Implement exporting

nmearl:
This implements basic exporting of data items. Since the io registry in specutils doesn't support exporting very well (re: at all, really), this also implements just a basic ECSV exporter as the one and only export format until the writers in specutils are fixed.


specviz/widgets/workspace.py
```diff
@@ -354,6 +356,69 @@ def compose_filter_string(reader):
...
+        all_filters = ";;".join(['*.ecsv'])
+
+        plot_data_item = self.current_item
+
+        try:
+            io_registry.register_writer('*.ecsv', Spectrum1D, generic_export)
+        except io_registry.IORegistryError:
+            pass
+
+        path, format = compat.getsavefilename(filters=all_filters)
```
drdavella: Just a minor comment, but `format` shadows a builtin.

specviz/widgets/workspace.py
```diff
@@ -354,6 +356,69 @@ def compose_filter_string(reader):
...
+        # all_filters = ";;".join(list(all_formats))
+
+        all_filters = ";;".join(['*.ecsv'])
+
+        plot_data_item = self.current_item
+
+        try:
+            io_registry.register_writer('*.ecsv', Spectrum1D, generic_export)
+        except io_registry.IORegistryError:
+            pass
```
drdavella: I wonder if it would make sense to have a warning/error here as well?

nmearl:
Fixed auto importing and added success dialog on export.

PR #593, opened 2018-11-29
Introduce simple model initializations

nmearl:
This re-introduces @ibusko's model initializer code. The initializer code in specutils needs to be built out more before it can be used here.


robelgeda:
Sorry about late update, this looks good to me.

PR #606, opened 2019-01-03
don't crash on new model with no selected item

gpernot:
just a missing "return"
also : added 'quit' menu entry


nmearl:
Thanks @gpernot! Could you look at the conflicts and rebase when you get a chance?

gpernot:
I messed up a bit my merge, but i think it's ok now.

nmearl:
Thanks! You'll have to get rid of the `astropy_helpers` changes before we can merge this. You should be able to do a simple `git checkout upstream/master -- astropy_helpers`.

Issue #609, opened 2019-01-24
Add spectrum selection dialog when opening file containing multiple spectra


drdavella:
Currently `specviz` can load data files that contain an arbitrary number of spectra (assuming that there is an associated loader in `specutils` for the file format that can parse all of the spectra in the file).

However, the current behavior is for `specviz` to populate *all* of the spectra in the file into the viewer. For files that contain large numbers of spectra, this can be burdensome for the user, and can also potentially cause performance degradation.

It would be nice if users could select the spectra that they wished to load from files that contain multiple spectra. To enable this, I propose adding a dialog that will appear only when opening such a file. The steps from a user perspective would look something like this:

1. Open the file loader dialog
2. Select the file to open, and click "open"
3a. If the file only contains a single spectrum, open the file and populate the data
3b. If the file contains multiple spectra, open another dialog that shows a list of all the spectra in the file.
4. By default, all spectra will be selected. The user can deselect any spectra they don't wish to display and click "load"

PR #612, opened 2019-01-25
Add tests for loading JWST data

### LINKED ISSUE #608: Add unit tests for JWST pipeline output data
We should probably add tests for all the following files, although we might be able to pair this down with some careful consideration:

NIRISS WFSS (Grism)  -  jw87600017001_02101_00002_nis_x1d_ref.fits 
NIRISS SOSS (Grism) - jw10003001002_03101_00001-seg003_nis_x1dints_ref.fits 
NIRSpec Fixed Slit - jw00023001001_01101_00001_NRS1_x1d_ref.fits
NIRSpec MSA (MOS) - f170lp-g235m_mos_observation-6-c0e0_001_dn_nrs1_mod_x1d_ref.fits
NIRSpec BOTS - jw84600042001_02101_00001_nrs2_x1dints_ref.fits
MIRI LRS Slitless time-series - jw80600012001_02101_00003_mirimage_x1dints_ref.fits
MIRI LRS fixed-slit time-series - jw00035001001_01101_00001_MIRIMAGE_x1dints_ref.fits
MIRI LRS fixed-slit - jw00035001001_01101_00001_MIRIMAGE_x1d_ref.fits

drdavella:
This resolves #608.

So obviously this took a lot less effort than I originally anticipated, but that's mostly because @SaOgaz had done all of the legwork of gathering the relevant test data and setting it up on Box.

Right now the test simply loads each data file into `specviz` and makes sure that nothing goes wrong. It's possible that we want to add more comprehensive checks to the tests, but I think this is actually a pretty good smoke test for now.

Each test case starts up a new instance of `specviz` in order to make sure we're simulating something like a real user workflow. In order to avoid the Qt segfault issue we discussed earlier in the week, each test case runs `specviz` in a separate process using the `multiprocessing` module. This might be worth using for other test cases as well.

### DIFF
--- a/specviz/tests/test_load_data.py
+++ b/specviz/tests/test_load_data.py
@@ -51,7 +51,7 @@ JWST_DATA_PATHS = [urljoin(BOX_PREFIX, name) for name in JWST_DATA_FILES]
 def load_jwst_data(url):
     try:
         spec_app = Application([], skip_splash=True)
-        data = spec_app.current_workspace.load_data(url, _interactive=False)
+        data = spec_app.current_workspace.load_data(url)
         # Basic sanity check to make sure there are data items
         assert len(data) > 0
     finally:
--- a/specviz/tests/test_load_data.py
+++ b/specviz/tests/test_load_data.py
@@ -50,7 +50,7 @@ JWST_DATA_PATHS = [urljoin(BOX_PREFIX, name) for name in JWST_DATA_FILES]
 def load_jwst_data(url):
     try:
         spec_app = Application([], skip_splash=True)
-        data = spec_app.current_workspace.load_data(url)
+        data = spec_app.current_workspace.load_data(url, _interactive=False)
         # Basic sanity check to make sure there are data items
         assert len(data) > 0
     finally:
--- a/specviz/tests/test_load_data.py
+++ b/specviz/tests/test_load_data.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import traceback
 from urllib.parse import urljoin
@@ -69,6 +70,10 @@ def run_specviz_subprocess(q, url):
     q.put(error)
 
 
+@pytest.mark.skipif(not os.environ.get('JWST_DATA_TEST'),
+                    reason='Since these tests run in a subprocess they do not '
+                    'play nicely with the fixture that is used for the rest of '
+                    'the test suite.')
 @pytest.mark.parametrize('url', JWST_DATA_PATHS)
 def test_load_jwst_data(url):
 
--- a/specviz/tests/test_load_data.py
+++ b/specviz/tests/test_load_data.py
@@ -89,3 +89,4 @@ def test_load_jwst_data(url):
         ex_type, ex_value, tb_str = error
         message = '{} (in subprocess)\n{}'.format(ex_value, tb_str)
         raise ex_type(message)
+
--- a/specviz/tests/test_load_data.py
+++ b/specviz/tests/test_load_data.py
@@ -1,0 +1,86 @@
+import sys
+import traceback
+from urllib.parse import urljoin
+from multiprocessing import Process, Queue
+
+import pytest
+
+from specviz.app import Application
+
+
+BOX_PREFIX = 'https://stsci.box.com/shared/static/'
+
+JWST_DATA_FILES = [
+    # NIRISS WFSS
+    # jw87600017001_02101_00002_nis_x1d_ref.fits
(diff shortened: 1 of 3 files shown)


specviz/tests/test_load_data.py
```diff
@@ -0,0 +1,86 @@
...
+
+JWST_DATA_PATHS = [urljoin(BOX_PREFIX, name) for name in JWST_DATA_FILES]
+
+
+def load_jwst_data(url):
+    try:
+        spec_app = Application([], skip_splash=True)
+        data = spec_app.current_workspace.load_data(url, _interactive=False)
+        # Basic sanity check to make sure there are data items
+        assert len(data) > 0
```
drdavella: This would be a little more effort, but we could check for the correct number of data items for each data file here.

specviz/widgets/workspace.py
```diff
@@ -512,15 +513,18 @@ def load_data(self, file_path, file_loader=None, display=False):
...
-            message_box = QMessageBox()
-            message_box.setText("Error loading data set.")
-            message_box.setIcon(QMessageBox.Critical)
-            message_box.setInformativeText(
-                "{}\n{}".format(
-                    sys.exc_info()[0], sys.exc_info()[1])
-            )
-
-            message_box.exec()
+            if _interactive:
```
nmearl: I'm not a _huge_ fan of the private method parameter here. Might it be better to move the qt-specific part of this method into the `_on_load_data` method (to keep those functions together), and reserve this one for the simple loading machinery?

drdavella: @nmearl yes, that makes sense. Some related refactoring is going to be required for my proposed feature in #609 as well.

nmearl:
Looks good!

drdavella:
**NOTE**: this shouldn't be merged quite yet: it looks like travis hooks aren't being delivered/received properly so the tests have not actually run yet.

drdavella:
Also, I want to make a note of the testing-related changes here.

As mentioned above, in order to avoid Qt-related issues when running multiple application instances in a single test process, each of the JWST data tests runs in a separate process by way of the `multiprocessing` module.

This appears to be very effective, but it still causes issues when these tests are run as part of the overall test suite. I haven't root-caused the issue, but there still appears to be some weird interaction with Qt, and also some problems with running Astropy across multiple processes.

In order to avoid this, I set up the JWST data loader tests to be skipped by default. When the test suite runs, these tests will be skipped, thus avoiding the problems. However, if the `JWST_DATA_TEST` environment variable is set, the tests will run.

So the travis test runs now actually run two separate test processes: one that runs the existing test suite and skips the JWST data tests, and another that runs only the JWST data tests. This is a bit of a kludgy workaround, but I don't think it's too problematic. We might want to revisit this at some point in the future.

specviz/tests/test_load_data.py
SaOgaz: This might be nice to have, but I'm going to approve this review even without that.

specviz/tests/test_load_data.py
```diff
@@ -0,0 +1,91 @@
...
+    # cause segfaults, so we run each specviz instance in a separate process
+    p = Process(target=run_specviz_subprocess, args=(q, url))
+    p.start()
+    error = q.get()
+    p.join()
+
+    if error:
+        ex_type, ex_value, tb_str = error
+        message = '{} (in subprocess)\n{}'.format(ex_value, tb_str)
+        raise ex_type(message)
```
SaOgaz: Blank line at the end of the file?

specviz/widgets/workspace.py
```diff
@@ -367,7 +367,18 @@ def compose_filter_string(reader):
...
+        except:
+            message_box = QMessageBox()
+            message_box.setText("Error loading data set.")
+            message_box.setIcon(QMessageBox.Critical)
+            message_box.setInformativeText(
+                "{}\n{}".format(
+                    sys.exc_info()[0], sys.exc_info()[1])
+            )
+
+            message_box.exec()
```
SaOgaz: Perhaps this is too nit picky @drdavella, but shouldn't this go into a different PR?  This is adding a nicer text box warning for failed file loads?

nmearl: This is not adding a message box, it's moving it from the `load_data` method to the `_on_load_data` signal callback.

drdavella:
Tests *should* pass once this is rebased on #614.

PR #614, opened 2019-01-28
Drop support for py35, update travis matrix

drdavella:
It appears that `specutils` will only be supporting Py36 and later going forward, and I believe that the same is true of the latest version of Astropy, so this seems like a prudent move. We were also encountering issues with Py35 on travis which seemed to indicate incompatible dependencies


drdavella:
Looks like require `astropy>=3.1` now due to the new WCS API.

PR #616, opened 2019-01-28
Make "Auto (*)" the default loader in open data dialog

### LINKED ISSUE #615: Should the loader selection dropdown default to "Auto"?
During a previous sprint I introduced a change to require users to explicitly choose a loader before clicking `Open` (see #491). This was to prevent the wrong option from being chosen inadvertently.

However, now that automatic loader selection has been implemented (#527), this no longer seems useful/necessary. Instead, it would make sense for the default to be `Auto`.

drdavella:
This fixes #615. It supersedes #491.

### DIFF
--- a/specviz/widgets/workspace.py
+++ b/specviz/widgets/workspace.py
@@ -351,13 +351,15 @@ class Workspace(QMainWindow):
 
         # Include an auto load function that lets the io machinery find the
         # most appropriate loader to use
-        loader_name_map['Auto (*)'] = None
+        auto_filter = 'Auto (*)'
+        loader_name_map[auto_filter] = None
 
-        # This ensures that users actively have to select a file type before
-        # being able to select a file. This should make it harder to
-        # accidentally load a file using the wrong type, which results in weird
-        # errors.
-        filters = ['Select loader...'] + list(loader_name_map.keys())
+        filters = list(loader_name_map.keys())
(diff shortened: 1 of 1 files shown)


drdavella:
Just want to mention here that it's not really practical to have a test for this functionality since it's just using a canned Qt dialog for finding files.

drdavella:
Tests won't pass until #614 is merged and this is rebased.

specviz/widgets/workspace.py
```diff
@@ -351,13 +351,15 @@ def compose_filter_string(reader):
...
-        filters = ['Select loader...'] + list(loader_name_map.keys())
+        auto_filter = 'Auto (*)'
+        loader_name_map[auto_filter] = None
+
+        filters = list(loader_name_map.keys())
+        # Make sure that the "Auto (*)" loader shows up first. Being a bit
+        # pedantic about this even though we can probably just rely on
+        # dictionary ordering here.
+        index = filters.index(auto_filter)
+        filters = filters[index:index+1] + filters[:index] + filters[index+1:]
```
nmearl: Just a bit of bikeshedding, but I find `filters = filters.insert(0, filters.pop(index))` a _smidge_ clearer. 🤓

drdavella: Doh 🤦‍♂️

PR #619, opened 2019-01-29
Fix failing tests; ensure specviz test instance sterility

nmearl:
Introduces handling of `QMessageBox` modals. Fixes failing tests in statistics plugin that required an unmodified workspace instance to sample for stats.


drdavella:
This is good to go, but now that #612 is merged it might be a good idea to do one more rebase and then change all calls to `setup.py test` to `pytest` in the travis config.

.travis.yml
```diff
@@ -139,7 +139,7 @@ script:
   # different test cases do not play nicely when they run in the same pytest
   # process, which is the reason for separating them here.
   - if [[ $CMD == test ]]; then
-      $SETUP_CMD test && JWST_DATA_TEST=1 $SETUP_CMD test -t $DATA_TEST_PATH;
+      pytest && JWST_DATA_TEST=1 pytest -t $DATA_TEST_PATH;
```
drdavella: `pytest` doesn't require the `-t` option, that's just for `setup.py test`

drdavella: Also, we actually need to do a `pip install -e .` now.

PR #622, opened 2019-01-29
Fix bug in load file dialog

drdavella:
This was a regression introduced by #616. This PR fixes the bug and adds a test to try to prevent it from cropping up again. However, this is a pretty tight corner between our application logic and Qt-specific components, so it's rather difficult to test properly.


nmearl:
I'm actually thinking that hitting these issues where the solution is "make a clear delineation between ui-logic and testable, core functionality" might be a good thing overall, if a bit tedious and frustrating to have to solve. This looks good!

drdavella:
@nmearl cool, there will be some more refactoring to that effect (out of necessity) in a forthcoming PR.

The problem is that the boundary will still be very difficult to test, but at some point we're just hitting on the limitations of the tools that we have.

drdavella:
Going ahead and merging this so the bug fix is available.

drdavella:
@kassin this should fix the issue on `master` that you reported.

PR #623, opened 2019-01-30
Add dialog for selecting which spectra to load

### LINKED ISSUE #609: Add spectrum selection dialog when opening file containing multiple spectra

drdavella:
This resolves #609. It provides a dialog that enables users to select which spectra to load from a file that contains multiple spectra.

This is pretty basic at the moment. We may want to consider adding a preview in the future. Also, it would eventually be useful for `specutils` loaders to provide more informative metadata about spectra (probably derived from the HDU headers) so that we could give more informative names to these data items.

The same open file dialog is used:
image

However, when loading a file with multiple spectra, users now see the following dialog after clicking "Open":
image

By default, all spectra are selected, so most users can just click "Open" and keep moving.

### DIFF
--- a/specviz/widgets/spectrum_selection.py
+++ b/specviz/widgets/spectrum_selection.py
@@ -14,7 +14,7 @@ class SpectrumSelection(QDialog):
             os.path.join(os.path.dirname(__file__),
                          "ui", "spectrum_selection.ui")), self)
 
-        self.setWindowTitle('Select Spectra to Load')
+        self.setWindowTitle('Spectrum Selection')
 
         self._model = QStandardItemModel(self.spectrumList)
         self.spectrumList.setModel(self._model)
--- a/specviz/widgets/spectrum_selection.py
+++ b/specviz/widgets/spectrum_selection.py
@@ -35,7 +35,7 @@ class SpectrumSelection(QDialog):
 
     def populate(self, names):
         """
-        Add a list of names to be displayed as list items in the dialog 
+        Add a list of names to be displayed as list items in the dialog
 
         Parameters
         ----------
--- a/specviz/widgets/spectrum_selection.py
+++ b/specviz/widgets/spectrum_selection.py
@@ -1,0 +1,69 @@
+import os
+
+from qtpy.uic import loadUi
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import QDialog
+from qtpy.QtGui import QStandardItemModel, QStandardItem
+
+
+class SpectrumSelection(QDialog):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        loadUi(os.path.abspath(
+            os.path.join(os.path.dirname(__file__),
+                         "ui", "spectrum_selection.ui")), self)
+
+        self.setWindowTitle('Select Spectra to Load')
+
+        self._model = QStandardItemModel(self.spectrumList)
+        self.spectrumList.setModel(self._model)
+        self._selected = False
+
+        self.buttonBox.accepted.connect(self._confirm_selection)
+
+        self.selectAllButton.clicked.connect(self._select_all)
+        self.deselectAllButton.clicked.connect(self._deselect_all)
+
+    def populate(self, spectra):
+        for s in spectra:
+            item = QStandardItem(s)
+            item.setCheckable(True)
+            item.setCheckState(Qt.Checked)
+            self._model.appendRow(item)
+
+    def get_selected(self):
+        if not self._selected:
+            return []
+
+        selected = []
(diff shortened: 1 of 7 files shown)


specviz/widgets/workspace.py
```diff
@@ -1,6 +1,7 @@
-import logging
 import os
 import sys
+import logging
+from collections import OrderedDict
```
nmearl: Dictionaries are ordered by default now.

specviz/widgets/workspace.py
```diff
@@ -357,25 +359,81 @@ def compose_filter_string(reader):
...
         return filters, loader_name_map
 
-    def _on_load_data(self):
-        """
-        When the user loads a data file, this method is triggered. It provides
-        a file open dialog and from the dialog attempts to create a new
-        :class:`~specutils.SpectrumList` object and thereafter adds the
-        contents to the data model.
-        """
+    def _choose_file_path(self):
```
nmearl: Docstrings.

specviz/widgets/workspace.py
```diff
@@ -357,25 +359,81 @@ def compose_filter_string(reader):
...
+
         filters, loader_name_map = self._create_loader_filters()
 
         file_path, fmt = compat.getopenfilename(parent=self,
                                                 basedir=os.getcwd(),
                                                 caption="Load spectral data file",
                                                 filters=";;".join(filters))
+        return file_path, loader_name_map[fmt]
+
+    def _load_spectra_by_name(self, specs_by_name):
```
nmearl: Docstring.

specviz/widgets/spectrum_selection.py
```diff
@@ -0,0 +1,69 @@
+import os
+
+from qtpy.uic import loadUi
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import QDialog
+from qtpy.QtGui import QStandardItemModel, QStandardItem
+
+
+class SpectrumSelection(QDialog):
```
nmearl: Docstrings.

specviz/widgets/spectrum_selection.py
```diff
@@ -0,0 +1,69 @@
...
+            item = self._model.item(i)
+            item.setCheckState(Qt.Checked)
+
+    def _deselect_all(self):
+        for i in range(self._model.rowCount()):
+            item = self._model.item(i)
+            item.setCheckState(Qt.Unchecked)
+
+
+if __name__ == '__main__': # noqa
```
nmearl: Is this being used for something outside of running the file locally?

specviz/widgets/workspace.py
drdavella: I'm being overly cautious here since we continue to support Py36, where that is only an implementation detail of CPython. It only became a language standard in Py37.

specviz/widgets/workspace.py
drdavella: So we need docstrings for all purely internal private methods as well?

specviz/widgets/spectrum_selection.py
drdavella: No, it's just very useful for development and debugging. I could have removed it but it's possible that someone will find it useful in the future.

drdavella:
@nmearl yes all data is parsed as a `SpectrumList` for consistent handling. This has been the case since #537 so there's nothing new in this PR.

specviz/widgets/workspace.py
nmearl: Ah, no, you're right. We don't need it there unless for explanation.

SaOgaz:
Update: all unit tests ran for me locally!

drdavella:
@SaOgaz it sounds like we were in agreement that the dialog resizing could be pushed off into a separate issue. If so, would you be willing to approve the rest of this PR?

SaOgaz:
Oh yeah sorry about that!

nmearl:
Once the rebase is done, we can merge!

PR #624, opened 2019-01-30
Add specutils version to splash screen

### LINKED ISSUE #613: Display specutils version on splash screen
This is a minor change and might be useful.

drdavella:
This resolves #613.

image


drdavella:
I guess it's worth asking whether we want to do this for `astropy` too, but that's probably not as important.

nmearl:
I would say no. If we really want that info, perhaps a pop-up under something like `File -> About` would be more appropriate.

PR #628, opened 2019-01-30
Line labels documentation

ibusko:
Somehow, the documentation page for the line labels plugin was lost. This PR adds it back.


nmearl:
@ibusko Once you fix the conflict, this should be good to go.

ibusko:
Can't understand the conflict. I even added the missing line in the file (the one with the 'developer' entry), and it still complains. The conflict report doesn't make sense to me. Merge should just stick the new line in master. Or is there something very basic that I am missing?

nmearl:
You can try using the GitHub editor to resolve the conflict. (Click the "Resolve Conflict" button). However, first, you're going to to need to get rid of all these merge commits with a rebase.

ibusko:
Interesting. Last time I tried the "Resolve conflicts" button (quite sometime ago) it would allow me to see the file, but not edit it. Now, the editor works. Glad you hinted me on that.

ibusko:
As for the rebase, I tried it this morning but nothing happened. I will try again.

ibusko:
Hmm..... I rebased, and force-pushed it, but the unwanted commits are still there.

ibusko:
Now it seems to have done something. Shouldn't the rebase remove all spurious commits and left just the few we want? It seems to have added the desired commits to the history, but without removing the previous commits. It had done that before, but I didn't notice, and tried again from the command line instead of PyCharm. Now the desired commits show up three times.

Maybe I am rebasing against the wrong branch. I use heads/master, the only one that makes sense and that PyCharm shows in its drop-down in the rebase dialog.

ibusko:
Yes, I also use upstream/master, but on the Pycharm rebase dialog, I have only entries like refs/heads/master for the 'Onto:' field. I see no references to upstream.

PR #639, opened 2019-01-31
Handle data loader errors consistently

drdavella:
This change allows data loader errors to be handled consistently (by displaying the error message in a pop-up dialog) both when loading from the GUI and when loading a file from the command line.

### DIFF
--- a/specviz/widgets/workspace.py
+++ b/specviz/widgets/workspace.py
@@ -357,6 +357,13 @@ class Workspace(QMainWindow):
 
         return filters, loader_name_map
 
+    def display_load_data_error(self, exp):
+        message_box = QMessageBox()
+        message_box.setText("Error loading data set.")
+        message_box.setIcon(QMessageBox.Critical)
+        message_box.setInformativeText(str(exp))
+        message_box.exec()
+
     def _on_load_data(self):
         """
         When the user loads a data file, this method is triggered. It provides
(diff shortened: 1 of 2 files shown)


specviz/widgets/workspace.py
```diff
@@ -357,6 +357,13 @@ def compose_filter_string(reader):
 
         return filters, loader_name_map
 
+    def display_load_data_error(self, exp):
+        message_box = QMessageBox()
```
nmearl: We may want to start using the `QMessageBox` class methods to generate dialogs, since they can be monkeypatched with qtbot for testing.

```python
    QMessageBox.critical(self, "Error loading data set.", str(exp))
```

drdavella: I can make that change here if you think it's appropriate.

nmearl: If it's not too much trouble, it'd save us in the future if the travis tests fail and we have run locally to figure out why (because travis won't tell you it's because it was hung up on an open dialog window).

drdavella: For some reason the title won't display when I set it up this way. 😕

nmearl: The title in the dialog box? (The "Error loading data set.")?

nmearl: We can skip making the change right now and explore it in the future.

PR #640, opened 2019-01-31
Make data loader handling more correct/robust

### LINKED ISSUE #618: Specviz crashes when loading an inappropriate data file
I accidentally just tried to load a JWST IFU data file and specviz crashed with the following:

```
specviz [INFO    ]: Spectrum file looks like wcs1d-fits
specviz [WARNING ]: Attempted load with 'wcs1d-fits' failed, trying next loader.
  File "/Users/ddavella/specviz/specviz/widgets/plotting.py", line 351, in _check_unit_compatibility
    plot_data_item = self.proxy_model.item_from_id(item.identifier)
  File "/Users/ddavella/specviz/specviz/core/models.py", line 133, in item_from_id
    self._items[data_item.identifier] = PlotDataItem(data_item)
  File "/Users/ddavella/specviz/specviz/core/items.py", line 75, in __init__
    self._spectral_axis_unit = self._data_item.spectral_axis.unit.to_string()
  File "/Users/ddavella/specviz/specviz/core/items.py", line 46, in spectral_axis
    return self.data(self.DataRole).spectral_axis
  File "/Users/ddavella/miniconda3/envs/specutils/lib/python3.7/site-packages/astropy/utils/decorators.py", line 726, in __get__
    val = self.fget(obj)
  File "/Users/ddavella/specutils/specutils/spectra/spectrum_mixin.py", line 63, in spectral_axis
    spectral_axis = self.wcs.pixel_to_world(np.arange(self.flux.shape[-1]))
AttributeError: 'NoneType' object has no attribute 'pixel_to_world'
```

This is probably actually a problem in the `specutils` loader, but it should be handled gracefully on this end.

drdavella:
This resolves #618. It is based on #639, which should be merged first. It also depends on a bug fix in `specutils` (see https://github.com/astropy/specutils/pull/440), so the tests will not pass until that is merged and released.

There were three fundamental problems here:
1. Most `specutils` loaders for `SpectrumList` were not providing identifiers. This was a bug and meant that if a user specified the wrong loader, even for a valid data file, the file would be opened but would likely cause a crash somewhere later. This is no longer possible with the `specutils` fix mentioned above.
2. In certain cases, `specviz` was not using identifiers to check whether a particular loader should be used before trying to attempt the file. Now that `specutils` provides identifiers for most cases, they can be used by `specutils` and an appropriate error can be propagated up to the user if no valid loader is available.
3. Prior to #639, loader errors were not being handled properly by `specviz` in all cases, so some loader errors would cause the application to crash (particularly when loading an invalid file from the command line).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
 
         # List other runtime dependencies for the package that are available as
         # pip packages here.
-        - PIP_DEPENDENCIES='pyqt5 pyqtgraph qtawesome qtpy click specutils>=0.5.1 asteval sphinx-astropy sphinx-rtd-theme pytest-astropy pytest-qt'
+        - PIP_DEPENDENCIES='pyqt5 pyqtgraph qtawesome qtpy click specutils>=0.5.2 asteval sphinx-astropy sphinx-rtd-theme pytest-astropy pytest-qt'
 
         # Conda packages for affiliated packages are hosted in channel
         # "astropy" while builds for astropy LTS with recent numpy versions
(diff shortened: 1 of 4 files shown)


drdavella:
**IMPORTANT**: Once `specutils` is released with the related PR, we need to update the minimum required `specutils` version here.

drdavella:
The CI run needs to be restarted once `specutils-0.5.2` is released.

drdavella:
It passes! 🎉

PR #644, opened 2019-02-01
Use fake data name for examples in arithmetic editor

### LINKED ISSUE #633: Equation Editor Examples should use simple file names
The example expressions given in the equation editor should used simplified file names.  These equations are very hard to read due to the complexity and length of the file names used.  Examples of good names are "spectrum1" and "spectrum2." 

screen shot 2019-01-30 at 3 05 58 pm

drdavella:
This resolves #633. Feel free to suggest a better name.

image


kassin:
spectrum1, spectrum2 are good fake names

drdavella:
Using `SPECTRUM1` and `SPECTRUM2`.

image

Issue #610, opened 2019-01-24
Create user facing documentation


javerbukh:
- Video tutorials
- Read through documentation and decide what needs to be done

javerbukh:
Includes finding scientists and user testing the documentation. This would be for quality assurance.

javerbukh:
@javerbukh will help to break this issue up and @kassin will help to find scientists. @javerbukh and @brechmos-stsci will be working on the video tutorial.

robelgeda:
My review of the SpecViz docs [Stable] :
-	SpecViz logo needs to be larger on the side bar.
-	Installation page looks very good:
	-	Can SpecViz run on python 2.x?
-	The subsections in the Getting Started section should be their own sections:
	-	Loading a data:
		-	List of supported spectra and file types
		-	How to load basic spectrum
		-	How to load spectrum from command line
		-	How to use the loader wizard [Provide quick tips or link to the Loader Wizard section]
		-	How to make a custom loader for unsupported data
	-	Exporting Spectra:
		-	As FITS?
		-	As ECSV 
	-	Workspaces and Plots:
		-	List all icons in on the toolbar and what they do
		-	How to create and remove new workspaces and plots 
	-	Using Regions:
		-	How to make and remove regions 
		-	Specify how the regions affect statistics and model fitting or link to the pages that do this
-	Unit Conversion section looks good.
-	Arithmetic Widget section looks good.
-	Loader Wizard section looks good.
	-	My personal openion is that this should be merged with a larger “how to load data section”
-	Improve Model fitting section by revising steps and adding screenshots.
-	Spectrum Statistics section looks good but we should indicate that the tabs on the side of specviz toggle b/w statistics and fitting.
-	Sections needed:
	-	Line labeling section. 
-	Proofread each page for typos. 
	-	On the index page (front page), remove the “SpecViz will soon include the ability to:” paragraph

@kassin @javerbukh

robelgeda:
Also, quick link to current docs:
http://specviz.readthedocs.io

astrofrog:
To install the latest-ish developer version with conda (built nightly):

```
conda install -c glueviz/label/dev specviz
```

I could potentially make the builds more frequent, and there might even be a way to trigger a build on a merge to master (but not sure)

kassin:
main page
---------

"SpecViz is built on top of the SpecUtils Astropy-affiliated python library, providing a visual, interactive interface to the analysis capabilities in that library."
->
[...]

PR #645, opened 2019-02-01
Improvements to User Docs

robelgeda:
This PR improves the user facing docs. Please see [this comment](https://github.com/spacetelescope/specviz/issues/610#issuecomment-459216717) for details on what is being introduced. 

For testing you can used this composite command:

`rm -r docs/_build/html; python setup.py build_docs; open docs/_build/html/index.html`

(You have to be at the top of your SpecViz clone where `setup.py` lives)

CC: @kassin @javerbukh


robelgeda:
Waiting on #628 to get merged

nluetzge:
Sorry I started commenting on the files itself as part of a review but then I ran into the issue on where to write comments about pages that you haven't uopdated so I'm collecting everything here. Let me know if there is a better way in the future.

Installation - Install via source:
- Hm maybe we should move the PyQt installation up, before the specviz installation? Or I guess it doesn't really matter? Would you still be able to install specviz without having PyQt already installed?

Launching SpecViz:
- Is there a way of getting a list of available loaders? Either in the cods or even make it a command line command so that the user can see what loaders are available before starting specviz? (Oh ok there is a list of loaders in the next section but its not 100% what I want)
- can we write an explicit example here. Because some of the loaders have spaces in their names and I realized that I need to put them in parenthesis in order to work from the commandline and it might be good to show the user to do this. 

Loader Wizard:
- maybe we should also mention how Specviz is currently selecting the loaders automatically. I believe this is still an issue that is not fully resolved yet but once we know we should share this information with the user. 
- also I think the future plan is to also suppoert FITS image data in the loader wizard
- this section probably needs a little re-write once the loader wizard is completed

Workspaces and Plots:
- The line labels are not in the plot toolbar anymore

Unit Conversion
- I always found this code snippet at the end a bit confusing. I think if we leave it in and I think its fine to leave it to see what it is doing we should describe the components in this equation (i.e. what is the spectral_axis_element?)

Arithmetic Editor:
- just a tiny little form thing. Can we make the screenshots nicer? With this I mean not to cut them out from a bigger screen shot but just making a dedicated screenshot of this window (so it gives you a nice shadow) like this:
screen shot 2019-02-04 at 6 51 53 pm

Model Fitting:
- That sounds a bit redundant : "The models can be fitted with the following fitters: They can be fit with the following fitting algorithms in astropy"

Other:
- we are still missing the line labels documentation.
- what about the continuum fitting? That's also still not in the docs. 

Otherwise I think they look great! Structure is good, layout is good. I think we soon need to ask somebody who has never used SpecViz before to see if the documentation is enough

robelgeda:
Thank you so much @nluetzge!

gderosa2004:
I agree with Nora that it looks good! I have several comments/suggestions for improvement below, but they can be implemented in the future.

- **Installation through anaconda**: since specviz requires python 3.6 I suggest to move the "create new conda environment with python 3.6 or higher higher in the text, after the conda info part. 

- **Installation, Pyqt**: add a few sentences about how can the user check if they have pyqt installed

-**Launching specviz**: here it is not clear what a "loader", can we give a definition and possibly add concrete examples?

-**Move the description of the "Loader wizard" before the "exporting spectra". Logically it belongs to the "launching" section**

- **Loader wizard**: I agree with Nora that this section needs a rewrite

- **Workspace Toolbar**:  typo "remobing" --> "removing"

- **Workspace toolbar**: Highlight in bold: "When applying an operation on a data item (fitting, removing etc.) you must first select the data from the data list before clicking the desired toolbar button."

- **Plots**:  "Any data that cannot be added to a plot will show as disabled in the data list." refer to plot above and give concrete examples

- **Plot toolba**r: typo "Rest the plot" --> "Reset the plot", also explain to what it gets reset

-**Unit conversion**: relevant pop-up windows difficult to read, they also get lost with all the rest of the info in the screenshots

- **Arithmetic editor**: Is it possible to divide two spectra? Explain workflow..

- **Model fitting**: what about continuum fitting?

**Overall**:
- screenshots should have filenames that are easy to read and to refer to (e.g. spectrum1, spectrum2) not our testing filenames.
- offer more examples with screenshots
- create use case scenarios and walk the user through the examples (e.g. fitting a line with a double gaussian in a continuum subtracted spectrum)
- what about errors on the spectra? I don't think I saw them mentioned anywhere..

robelgeda:
@nluetzge Thanks for the awesome review, I checked everything I changed. Here are some things I did/could not change and why:

> Loader Wizard

The loader wizard was moved to a loading data section. I think it is still evolving so we are going to provide very general instructions for now.

> * just a tiny little form thing. Can we make the screenshots nicer? With this I mean not to cut them out from a bigger screen shot but just making a dedicated screenshot of this window (so it gives you a nice shadow) like this:

This is awesome!!! Since we are on a time crunch I am going to open a ticket for this and we will address this on its own.

> * what about the continuum fitting? That's also still not in the docs.

I don't think this has been implemented yet: cc @nmearl

robelgeda:
@gderosa2004 Thank you so much for the insights! 

I checked everything I changed. Here are some things I did/could not change and why:

> * **Loader wizard**: I agree with Nora that this section needs a rewrite

Same as above, the loader wizard was moved to the loading data section. I think it is still evolving so we are going to provide very general instructions for now.

> * **Plots**:  "Any data that cannot be added to a plot will show as disabled in the data list." refer to plot above and give concrete examples

> * **Unit conversion**: relevant pop-up windows difficult to read, they also get lost with all the rest of the info in the screenshots

> * screenshots should have filenames that are easy to read and to refer to (e.g. spectrum1, spectrum2) not our testing filenames.

> * offer more examples with screenshots

I will create a separate ticket (#664) to address this. Given the effort and time we need for this I think its best to tackle it in another PR / sprint. 

> * create use case scenarios and walk the user through the examples (e.g. fitting a line with a double gaussian in a continuum subtracted spectrum)

> * **Arithmetic editor**: Is it possible to divide two spectra? Explain workflow..

We are going to make video tutorials and I think we can include these tasks in them. 

> * **Model fitting**: what about continuum fitting?

I don't think this has been implemented yet. Please see above.

> * what about errors on the spectra? I don't think I saw them mentioned anywhere..

Good point. I need to look into how this works and will open a ticket for this (#665)

robelgeda:
@nluetzge @gderosa2004: I pushed my changes.

gderosa2004:
@robelgeda updates following my previous comments look good. I went quickly through the loader wizard and line list sections and they look ok for the time being. I didn't read through the last section though. Did anyone look into it?

robelgeda:
The last section was written this sprint and introduced in another PR so I think its good.

robelgeda:
@nluetzge 's review:

- [x] Model Fitting: "The models can be fitted with the following fitters: They can be fit with the following fitting algorithms in astropy"

- [x]  Linelists in the sidebar

- [ ] Current images of the toolbar should be replaced because the new one doesn't have the line lists in that toolbar anymore

PR #648, opened 2019-02-04
don't crash on new model with no selected item

gpernot:
i reopen this pull request as i messed up to much in previous merge :/


nmearl:
Thanks @gpernot!

PR #646, opened 2019-02-02
Line labels plugin in the plugin bar.

ibusko:
This PR replaces the previous one with same code. The goal is to have the line labels plugin running in the plugin bar, instead as a floating window. 

This PR was created from a freshsly created branch in my fork, using as base the previous plugin_linelabels branch. This was after many unsucessful atempts to rebase, which I was afraid could compromise the branch and its PR somehow. Apparently there are too many commits with conflicts, that cause the rebase to choke. A possible solution for this would be to squash just the commits with conflicts, one by one, by hand. I may still try this, but I want to have a PR with working code in place here. 

Eventually we managed to get the conflicts resolved one by one (there were dozens...)

The code in here already contains the changes from @nmearl 's code review.

PR #601, opened 2018-12-17
Line labels code refactored as plug-in, using the plug-in bar

ibusko:
This PR replaces the old one (#553) which kept the plug-in in a floating, separate window. In this one, the line labels controls are all kept confined inside the plug-in bar.

PR #650, opened 2019-02-04
Line labels plugin runs in plugin bar

ibusko:
This is a cleaner version of the older PR #646. The commit history was re-made into a single comit, so we could get rid of the many non-significant commits (resulting from merges from upstream) that refused to go away when rebasing.

I suspect this was caused by something broken in the fork where #646 comes from, because other PRs from that fork also showed the same behavior. 

The code in here is a plain copy of the same code that was already reviewed in #646. Please refer to #646 to access the code review and previous conversation. The original, first PR in this series, is #601.


nmearl:
To get rid of some of the docs build errors, be sure to remove the module references in the `specviz/docs/api.rst`. For instance, `specviz.core.annotation` can be removed, along with some others.

ibusko:
@nmearl It seems I got rid of the doc build errors. The remaining build errors are more difficult to figure out. 

For instance, the build_docs error apparently is about icon files that can't be found, even though they are in the resources directory. I wonder if that has to do with the new Label-48.png icon in that directory. I found that pytest-qt can't handle png icons. Seems unrelated, but maybe something similar is happening here.

The other two build tests end in core dumps. It's hard to track withou access to the CI environment. Do you have any suggestions?

ibusko:
@nmearl I don't think I ever set up the proper environment to run CI tests. How do I set it up?

nmearl:
Run `pytest specviz` in your `specviz` directory.

ibusko:
I always get an empty run terminating with Abort trap: 6

```
(specviz_dev) bash-3.2$ pytest specviz_stscissb
======================================= test session starts =======================================
platform darwin -- Python 3.7.0, pytest-3.5.1, py-1.7.0, pluggy-0.6.0
PyQt5 5.11.2 -- Qt runtime 5.11.1 -- Qt compiled 5.11.1
rootdir: /Users/busko/Projects/specviz_stscissb, inifile: setup.cfg
plugins: remotedata-0.3.0, qt-3.0.0, openfiles-0.3.0, doctestplus-0.1.3, arraydiff-0.2
collecting 32 items / 3 errors                                                                    Abort trap: 6
(specviz_dev) bash-3.2$ 
```

nmearl:
You may need to fix the icon issue first.

ibusko:
When fixing the icon, I get this:

```
(specviz_dev) bash-3.2$ pyrcc5 specviz/data/resources/resources.qrc -o specviz/widgets/resources.py
Cannot find file: specviz.icns
(specviz_dev) bash-3.2$ 

```

nmearl:
Weird, add this file to your `specviz/data/resources` directory: https://stsci.box.com/s/6nd6kkjk3ufmxbi8w2y566a8vkqqh0fv

ibusko:
Box says the file was removed or is not available.

nmearl:
Try the link now.

nmearl:
When I use a `.svg` file instead of the `.png`, all tests pass. I'm not sure what the issue is about, but I suggest you just use this icon: https://stsci.box.com/s/7ocy6ry87ls7tvjo1hyzvyalbmv0uxpw

ibusko:
That's weird. Thanks for finding this. At least now we have a price tag and can charge cash for specviz :-)

Now there is only one doc test still failing. It fails because warnings are turned into errors by the -W flag in sphinx. When I run it locally with no -W flag, it passes. Although there are 392 warnings, all associated to the plotting.py module, as in "toctree contains reference to nonexisting document 'api/specviz.widgets.plotting.QMdiSubWindow". 

The end product of that run (the html page) suggests that sphinx is attempting to treat Qt classes as specviz classes because it shows some of the Qt classes in the output.

nmearl:
😆 

> It fails because warnings are turned into errors by the -W flag in sphinx.

This is intentional. Warnings indicate that the docs were not built or not built correctly and are intended to block merges.

Since the plotting module api docs built before this PR, it seems we're still introducing something weird in this case. Paging @astrofrog.

specviz/plugins/line_labels/annotation.py
```diff
@@ -0,0 +1,186 @@
...
+
+from qtpy.QtCore import QPointF, QRectF
+from qtpy.QtGui import QPolygonF, QPen, QColor
+
+orientations = {
+    'horizontal': {'anchor': (0.5, 1), 'angle': 0},
+    'vertical': {'anchor': (0, 0.5), 'angle': 90}
+}
+
+
```
astrofrog: This file is missing an ``__all__``

specviz/plugins/line_labels/line_labels_plotter.py
```diff
@@ -1,89 +1,135 @@
...
+from astropy import units as u
 
-__all__ = ['LineLabelsPlotter', 'ZoomMarkersThread', 'ZoomEventBuffer']
+from .annotation import LineIDMarker, LineIDMarkerProxy
+from .linelist import LineList, WAVELENGTH_COLUMN, \
+    REDSHIFTED_WAVELENGTH_COLUMN, \
+    MARKER_COLUMN, ID_COLUMN, COLOR_COLUMN, HEIGHT_COLUMN
 
 
 class LineLabelsPlotter(object):
```
astrofrog: This file is also missing an ``__all__``

specviz/plugins/line_labels/linelists_window.py
```diff
@@ -47,9 +45,20 @@
 PLOTTED = "Plotted"
```
astrofrog: Missing ``__all__``

specviz/widgets/plotting.py
```diff
@@ -17,18 +17,16 @@
...
-from ..core.linelist import LineList, WAVELENGTH_COLUMN, ID_COLUMN
-from .line_labels_plotter import LineLabelsPlotter
 
 __all__ = ['PlotWindow', 'PlotWidget']
 
 
 class PlotWindow(QMdiSubWindow):
     """
-    Displayed plotting subwindow available in the QMdiArea.
+    Displayed plotting subwindow available in the `QMdiArea`.
```
astrofrog: Don't add backticks here - this is what's causing the error. Single backticks mean a proper API reference. Use double backticks if you want it to look like code.

ibusko: Done.

specviz/plugins/line_labels/linelists_window.py
ibusko: Done.

specviz/plugins/line_labels/line_labels_plotter.py
ibusko: Done.

specviz/plugins/line_labels/annotation.py
ibusko: Done.

PR #666, opened 2019-02-07
remember path location between load spectrum dialogs

gpernot:


specviz/widgets/workspace.py
```diff
@@ -122,6 +122,9 @@ def __init__(self, *args, **kwargs):
         self._plugins = {}
         self._plugin_bars = {}
 
+        # Current working directory. Updated by file open dialogs
+        self.cwd = os.getcwd()
```
nmearl: Just a small quibble, but would you mind making this a private variable? E.g. `self._cwd`?

gpernot: Ok. See latest commit

Issue #602, opened 2019-01-02
Continuum fitting crashes if there is no ROI


hcferguson:
Continuum crashes if there is no ROI.

```Traceback (most recent call last):
  File "/Users/hcferguson/anaconda3/envs/specutils_dev/lib/python3.6/site-packages/specviz-0.6.dev1898-py3.6.egg/specviz/core/plugin.py", line 176, in <lambda>
    action.triggered.connect(lambda: func(plugin, *args, **kwargs))
  File "/Users/hcferguson/anaconda3/envs/specutils_dev/lib/python3.6/site-packages/specviz-0.6.dev1898-py3.6.egg/specviz/plugins/model_editor/continuum_generator.py", line 32, in on_action_triggered
    exc_regs = inc_regs.invert_from_spectrum(spec)
AttributeError: 'NoneType' object has no attribute 'invert_from_spectrum'
[1]    96682 abort      specviz
```

PR #669, opened 2019-02-08
Fix ROIs in model generation; fixes #604 and #602

### LINKED ISSUE #604: Model creation crashes when there are multiple ROI
Working with stis1.ecsv, I defined 4 ROIs then tried to create a 2nd-order polynomial model. SpecViz crashed with the following traceback.

```
  File "/Users/hcferguson/anaconda3/envs/specutils_dev/lib/python3.6/site-packages/specviz-0.6.dev1898-py3.6.egg/specviz/plugins/model_editor/model_editor.py", line 68, in <lambda>
    action.triggered.connect(lambda x, m=v: self._add_fittable_model(m))
  File "/Users/hcferguson/anaconda3/envs/specutils_dev/lib/python3.6/site-packages/specviz-0.6.dev1898-py3.6.egg/specviz/plugins/model_editor/model_editor.py", line 263, in _add_fittable_model
    model = initialize(model, spec.spectral_axis, spec.flux)
AttributeError: 'list' object has no attribute 'spectral_axis'
[1]    97936 abort      specviz
```

nmearl:
FIxes #604 and #602.

### DIFF
--- a/specviz/plugins/model_editor/model_editor.py
+++ b/specviz/plugins/model_editor/model_editor.py
@@ -10,7 +10,7 @@ from qtpy.QtWidgets import (QAction, QDialog, QFileDialog, QInputDialog, QMenu,
                             QMessageBox, QToolButton, QWidget)
 from qtpy.uic import loadUi
 from specutils.fitting import fit_lines
-from specutils.manipulation import extract_region
+from specutils.manipulation.utils import excise_regions
 from specutils.spectra import Spectrum1D
 
 from .equation_editor_dialog import ModelEquationEditorDialog
@@ -259,7 +259,10 @@ class ModelEditor(QWidget):
         spec = self._get_selected_plot_data_item().data_item.spectrum
 
         if inc_regs is not None:
-            spec = extract_region(spec, inc_regs)
+            exc_regs = inc_regs.invert_from_spectrum(spec) \
+                if inc_regs is not None else None
+
+            spec = excise_regions(spec, exc_regs)
 
         # Initialize the parameters
         model = initialize(model, spec.spectral_axis, spec.flux)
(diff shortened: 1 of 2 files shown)


specviz/plugins/model_editor/model_editor.py
```diff
@@ -259,7 +259,10 @@ def _add_fittable_model(self, model_type):
         spec = self._get_selected_plot_data_item().data_item.spectrum
 
         if inc_regs is not None:
-            spec = extract_region(spec, inc_regs)
+            exc_regs = inc_regs.invert_from_spectrum(spec) \
+                if inc_regs is not None else None
```
brechmos-stsci: For consistency this seems to be the same structure as above change, so how about just:
```
       exc_regs = inc_regs.invert_from_spectrum(spec) \
            if inc_regs is not None else None
```
though maybe I am missing something.

brechmos-stsci: nvm, missed the line below.

brechmos-stsci:
I think I need a bit of guidance here. I am not sure how to setup in order to get to that bit of code. I loaded a 1D data file, created two regions (hoping to use those to exclude for model fitting), then created a 2nd order polynomial. The fitting works (but it is a bad fit). How do I specify that these two regions I created should be used to exclude?  

Or quite possibly I am missing something with what @hcferguson  was doing in https://github.com/spacetelescope/specviz/issues/604 and what this fix is for.

This does appear to fix https://github.com/spacetelescope/specviz/issues/602.

nmearl:
The ROIs represent regions to be included. The code retrieves the current ROI select (`inc_regs`), and then inverts them to get the non-ROI selection of the spectrum (`exc_regs`), and feeds that into the specutils utilities (since they take excluded regions, not included regions). Hope that makes sense!

brechmos-stsci:
Cool. Seems to work for me, then.

PR #670, opened 2019-02-11
Use tox for testing

drdavella:
The motivation for this is that it allows for push-button replication of CI test environments locally. In order to run the tests using `tox` locally, it is necessary to have a `conda` environment with both `tox` and the `tox-conda` plugin installed (which are both currently only available from `pypi`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ matrix:
 
     include:
         # Do a coverage test
-        - env: TOXENV='coverage' TOX_ARGS=''
+        # Skip for now. Eventually we will want to use codecov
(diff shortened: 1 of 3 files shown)


drdavella:
Basically the only problem now is that coverage is not set up for this repository. We can either set it up now or just disable that job for the time being.

nmearl:
Seems like the test in question uses coveralls. If all the other viz tools are going for codecov, I think we should do that too. For now, though, we can disable it. I'm unsure how difficult it is to setup with codecov, and it's really not necessary during this current sprint.

Issue #679, opened 2019-03-01
Change data passed into `add_component` to be a glue Component to attach units


brechmos-stsci:
From @nmearl in https://github.com/spacetelescope/cubeviz/issues/501#issuecomment-468693750:

> The data passed to the `add_component` method is a simple array-like. This needs to be changed to an explicitly glue `Component` object in order to attached associated units. This would need to be done in specviz's glue data viewer.

nmearl:
Fixed with #680.

PR #680, opened 2019-03-01
Glue data viewer improvements

nmearl:
This PR fixes:
- Propagation of unit information (#679) 
- Deletion of plotted items on subset deletion


nmearl:
Thanks for keeping me honest, @drdavella. 😆 Added test.

brechmos-stsci:
I'll take a look at the unit propagation

brechmos-stsci:
I tested on a dataset with units "ADU' and they were passed back in the spectrally smoothed dataset. So, that correction looks good.

Issue #684, opened 2019-03-05
Median smoothing dialog width can not be interpreted as an integer


brechmos-stsci:
Tried doing a median filter with "3" as the width and got an error:
![image](https://user-images.githubusercontent.com/18348847/53820532-f5246680-3f39-11e9-85e7-b2b5c4eec0fb.png)

This was based on the SpecViz plugin in CubeViz.

Issue #683, opened 2019-03-05
Changing units via the third party plugin causes exception


brechmos-stsci:
I loaded the manga data into cubeviz, changed the units to micrometer (from the value of meter that from when it was loaded).  Then switched back to meter and got the following exception:

```
Traceback (most recent call last):
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/cubeviz/tools/wavelengths_ui.py", line 98, in _calculate_callback
    self.do_calculation(wavelength_redshift=redshift, wavelength_units=units)                                               
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/cubeviz/tools/wavelengths_ui.py", line 70, in do_calculation
    self.wavelength_controller.update_units(wavelength_units)
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/cubeviz/controls/wavelengths.py", line 84, in update_units
    self._send_wavelength_unit_message(units)
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/cubeviz/controls/wavelengths.py", line 113, in _send_wavelength_unit_message
    self._hub.broadcast(msg)
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/glue/core/hub.py", line 217, in broadcast           
    handler(message)
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/cubeviz/image_viewer.py", line 245, in <lambda>     
    handler=lambda msg: self.cubeviz_layout.specviz._widget.update_units(spectral_axis_unit=msg.units))                     
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/specviz/third_party/glue/viewer.py", line 452, in update_units
    self.hub.plot_widget.spectral_axis_unit = spectral_axis_unit                                                            
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/specviz/widgets/plotting.py", line 304, in spectral_axis_unit
    plot_data_item.spectral_axis_unit = value
  File "/opt/miniconda3/envs/cubeviz-v0.3b1/lib/python3.6/site-packages/specviz/core/items.py", line 271, in spectral_axis_unit
[...]

Issue #667, opened 2019-02-07
Different symbols for "Loader Wizard", "New Model" and "Change Units"


nluetzge:
Currently the those three above mentioned functions have the same symbol:
<img width="23" alt="screen shot 2019-02-07 at 5 45 56 pm" src="https://user-images.githubusercontent.com/24398824/52447922-43437700-2b00-11e9-8b7f-aa50014babc4.png">
Can we make them different? I find it confusing that they are the same.

Issue #636, opened 2019-01-30
"Export Plot" not working


kassin:
Do not show this option if it is not yet implemented.  Grey it out or remove it from the toolbar.

drdavella:
What is the intent behind this? It looks like `pyqtgraph` provides a few different formats for exporting plots, including JPEG, CSV, matplotlib, and others. Do we want to expose a few different options to the user, or is there one right answer here?

PR #689, opened 2019-03-08
Improvements and bug fixes

### LINKED ISSUE #690: specviz simple linemap incorrect?
I'm not sure if I'm understanding what was intended for the `simple linemap`, but I'm not getting what I expected. 

My understanding is that you select your wavelength range using ROIs on the 1D spectrum, and then choose `Cube Operations -> simple linemap` to sum over this wavelength region ** for every spaxel in the data cube, ** (regardless of whether the spectrum itself was selected from a subset). I expected the result to be a 2D image. What I'm getting back is a 3D image.

Looking at the code, it's not instantly obvious to me how it's supposed to work but (a) it looks like it is making a 3D image, and (b) I don't understand why it is doing this in a loop. If the mask is 0 where you should ignore and 1 where you should keep the data, then the line map is just `(data*mask).sum(axis=0)` assuming the spectral axis is axis 0.

Existing code in viewer.py:
```
    def _create_simple_linemap(self):
        def threadable_function(data, tracker):
            out = np.empty(shape=data.shape)
            mask = self.hub.region_mask

            for x in range(data.shape[1]):
                for y in range(data.shape[2]):
                    out[:, x, y] = np.sum(data[:, x, y][mask])
                    tracker()

            return out, data.meta.get('unit')
```

nmearl:
- Fixes interop unit change issues with cubeviz #683 
- Enables plot exporting #636 
- Fixes median smoothing issues #684 
- Line widths can now be changed. Line widths > 1 auto-enable openGL rendering to avoid performance issues. #631 
- ROIs update in response to unit changes #626
- Make ui icons unique #667 
- Introduces box zooming toggle option #321 
- Linemap operations now generate overlays instead of components. Fixes #690.
- Fixes fitted linemaps #691 
- Adds new cube operation to generate 3d components from fitted models #692 
- Fixes arithmetic issue when names have colons #693


specviz/plugins/arithmetic/arithmetic_editor.py
```diff
@@ -245,10 +250,10 @@ def _update_status(self):
...
                 self.label_status.setText("Incomplete or invalid syntax")
                 self.button_ok.setEnabled(False)
-            except Exception as exc:
-                self.label_status.setStyleSheet('color: red')
-                self.label_status.setText(str(exc))
-                self.button_ok.setEnabled(False)
+            # except Exception as exc:
+            #     self.label_status.setStyleSheet('color: red')
+            #     self.label_status.setText(str(exc))
+            #     self.button_ok.setEnabled(False)
```
brechmos-stsci: if these aren't wanted here, probably best to just delete the code.

specviz/plugins/smoothing/smoothing_dialog.py
```diff
@@ -174,7 +174,7 @@ def accept(self):
         self.smooth_button.setEnabled(False)
         self.cancel_button.setEnabled(False)
 
-        self.size = float(self.size_input.text())
+        self.size = int(self.size_input.text())
```
brechmos-stsci: I wonder if an error/warning might be better here or a check to see `int()` is close to desired.  if the person enters `3.95` then this will truncate the value to `3` which is quite a ways off.

nmearl: The field validator on the line edit will ensure the user cannot enter a float.

specviz/plugins/arithmetic/arithmetic_editor.py
nmearl: You're right; took them out for debugging and never uncommented. Fixed.

specviz/plugins/smoothing/smoothing_dialog.py
```diff
@@ -80,6 +81,9 @@ def __init__(self, parent=None, *args, **kwargs):
             self.kernel_combo.addItem(kernel["name"], key)
         self.kernel_combo.currentIndexChanged.connect(self._on_kernel_change)
 
+        # Add integer validator to size input field
+        self.size_input.setValidator(QIntValidator())
```
drdavella: This is cool. We should be doing this everywhere in cubeviz.

PR #702, opened 2020-10-09
Added note pointing to Jdaviz as the maintained repo

rosteen:
Directed users to Jdaviz in case they end up here via e.g. Google searching

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,10 @@
+Current Status
+--------------
+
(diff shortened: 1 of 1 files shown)


dtn5ah (approved):
Thanks Ricky!

javerbukh (approved):
Looks good.

javerbukh:
Merging.

